PR #17, opened 2020-06-18
Apple/storing change vaildator

HarleyAppleChoi:
Storing entry changed by MsgEditValidator


schema/staking.sql
```diff
@@ -43,3 +43,14 @@ CREATE TABLE validator_unbonding_delegations
...
+CREATE TABLE validator_commission
+(
+    validator_address     CHARACTER VARYING(52) NOT NULL REFERENCES validator (consensus_address),
+    timestamp             TIMESTAMP WITHOUT TIME ZONE NOT NULL,
+    commission            decimal	              NOT NULL,
+    min_self_delegation    BIGINT              NOT NULL,
+    Discription          CHARACTER VARYING(52) NOT NULL,
+    height                BIGINT                NOT NULL,
+    PRIMARY KEY (validator_address,height)
+);
```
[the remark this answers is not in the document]
HarleyAppleChoi: Since the message is passed, we don't know whether is the commission changed or min self delegation/description changed. If the user changed the self delegation/description but not the commission rate, there are duplicated entries. So I think that's better to store the entries that can be changed by the MsgEditValidator. Though changing the name of the table should be a good idea?

x/staking/handler.go
```diff
@@ -0,0 +1,54 @@
...
+	if found, _ := db.HasValidator(address.String()); !found {
+		return nil
+	}
+
+	var validator staking.Validator
+	endpoint := fmt.Sprintf("/staking/validators/$s", address.String())
+	height, ok := cp.QueryLCDWithHeight(endpoint, &validator)
+	if ok != nil {
+		return ok
+	}
```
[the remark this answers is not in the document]
HarleyAppleChoi: When should I take the message from LCD and from msg?
Kwan said I should take from LCD when getting the latest state of the chain.

schema/staking.sql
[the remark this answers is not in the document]
RiccardoM: > Since the message is passed, we don't know whether is the commission changed or min self delegation/description changed. If the user changed the self delegation/description but not the commission rate, there are duplicated entries. 

You're right. I didn't think about it. I only got two observations about these points: 

1. We might not be interested in storing the historical changes of the validator description.  
   Usually the only data that is relevant for explorers is the latest description available. For this reason, I would just store it inside another table that does not change that frequenly and has at most one row per validator (this can be the `validator_info` table adding a column to it). 

2. To solve the duplication issue, what we might do is actually querying the database before saving the information.  
    If we just query the latest row of the database abased on the associated height and check if the commission rate is different from the one present inside `MsgEditValidator`, then we can decide to skip the storing. This would avoid duplicate entries when something other than the current commission rate value changes.

x/staking/handler.go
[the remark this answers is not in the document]
RiccardoM: > When should I take the message from LCD and from msg?

Since the LCD has some overhead associated to it, if possible always use the data present inside the message. 

Messages inside transactions are always parsed, even if you do not use them, cause they are also sent to other 
`MsgHandler` methods, so registering a new `MsgHandler` and using the data inside the messages has ideally no overhead at all. 

On the other hand, querying an LCD means that a network request must be done and the response must be de-serialized twice: one to get the response and another one to get its content.

---

> Kwan said I should take from LCD when getting the latest state of the chain.

LCD requests should be done only when querying the data that cannot be queried in any other way (this is the case of data that is not changed thought messages such as the staking pool or the staking rewards), and not when the same data can be take from blocks/transactions/messages/

HarleyAppleChoi:
Please have a look and see it is good or merge or not?

x/staking/handler.go
```diff
@@ -0,0 +1,60 @@
...
+			Msg("skipping message as it was not successful")
+		return nil
+	}
+
+	db, ok := w.Db.(database.BigDipperDb)
+	if !ok {
+		return fmt.Errorf("given database instance is not a BigDipperDb")
+	}
+
+	timestamp, err := time.Parse("2006-01-02T15:04:05Z", tx.Timestamp)
```
RiccardoM, line 32: This time format can be simplified by using 

```suggestion
	timestamp, err := time.Parse(time.RFC3339, tx.Timestamp)
```

x/staking/handler.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	if len(tx.Logs) == 0 {
+		log.Info().Msg(fmt.Sprintf("Skipping message at index %d of tx hash %s as it was not successull",
+			index, tx.TxHash))
+		return nil
+	}
+	switch stakingMsg := msg.(type) {
+	case staking.MsgEditValidator:
+		// TODO: Handle message here
+		//store commission rate
```
RiccardoM, line 44: I would delete these comments

```suggestion
```

x/staking/handler.go
```diff
@@ -0,0 +1,60 @@
...
+		StoreEditValidator(stakingMsg, w.ClientProxy, timestamp, tx.Height, db)
+	}
+
+	return nil
+}
+
+func StoreEditValidator(msg staking.MsgEditValidator, cp client.ClientProxy, time time.Time, height int64, db database.BigDipperDb) error {
+	//should I take from REST or store the message?
+	//store the message
```
RiccardoM, line 53:
```suggestion
```

x/staking/handler.go
```diff
@@ -0,0 +1,60 @@
...
+	address := msg.ValidatorAddress
+	if found, _ := db.HasValidator(address.String()); !found {
+		return nil
+	}
```
RiccardoM, line 57: I would delete this check and let the `Database` class handle it

database/staking.go
```diff
@@ -99,15 +117,62 @@ func (db BigDipperDb) SaveValidators(validators []bstaking.Validator) error {
...
+		return err1
+	}
+	for rows.Next() {
+		err := rows.Scan(&c, &m, &(d[0]), &(d[1]), &(d[2]), &(d[3]), &(d[4]))
+		if err != nil {
+			return err
+		}
+		if commissionRate == c || minSelfDelegation == m {
+			//commission rate changed(insert)
+			statement := `INSERT INTO validator_commission (validatorAddress,commissions,min_self_delegtion,height,timestamp) VALUES ($1,$2,$3,$4,$5);`
```
RiccardoM, line 152: I think this query would run return an error since the `validatorAddress` column does not exist. 

We should change it to be 

```suggestion
			statement := `INSERT INTO validator_commission (validator_address,commissions,min_self_delegtion,height,timestamp) VALUES ($1,$2,$3,$4,$5);`
```

database/staking.go
```diff
@@ -99,15 +117,62 @@ func (db BigDipperDb) SaveValidators(validators []bstaking.Validator) error {
...
-// SaveValidatorUptime stores into the database the given validator uptime information.
-// It assumes that for each uptime information provided, the associated validator data
-// have already been saved inside the database properly.
-func (db BigDipperDb) SaveValidatorUptime(uptime bstaking.ValidatorUptime) error {
-	statement := `INSERT INTO validator_uptime (height, validator_address, signed_blocks_window, missed_blocks_counter)
-				  VALUES ($1, $2, $3, $4)`
-	_, err := db.Sql.Exec(statement,
-		uptime.Height, uptime.ValidatorAddress.String(), uptime.SignedBlocksWindow, uptime.MissedBlocksCounter)
-	return err
+func (db BigDipperDb) SaveEditValidator(validator sdk.ValAddress, commissionRate int64, minSelfDelegation int64,
```
RiccardoM, line 120: As far as I've seen, this method performs the following operations: 

- check if the validator exists
- get the latest validator info and commission
- edit the validator info and commission

I personally have some suggestions on how the implementation of this method might be made more easier to understand: 

1. Split this method into smaller methods.
   - To get the latest validator info create a new `GetValidatorInfo` method that accepts the consensus address and returns the validator info data
   - To get the latest validator commission create a new `GetValidatorCommission` method that accepts the consensus address and returns the latest commission rate for such validator
   - To store the validator info, create a new `SaveValidatorInfo` method 
   - To store the validator commission, create a new `SaveValidatorCommission` method 

2. Create specific types for each table.  
   Since we are dealing with a lot of tables, it would be much easier to have custom types for each of them in order to represent a single entry inside them.  
   An example could be:  
   ```golang
   type ValidatorCommission struct {
     ValidatorAddress string `db:"validator_address"`
     Timestamp time.Time `db:"timestamp"`
     Commission float64 `db"commission"`
     MinSelfDelegation int64 `db:"min_self_delegation"`
     Height int64 `db:"height"`
   }
   ```
   This would make it much easier for you to both insert and query data using [`Sqlx` support to query and insert data using custom struct definitions](https://stackoverflow.com/a/40884611/5162161) ( an example inside BDJuno can be found [here](https://github.com/forbole/bdjuno/blob/master/database/staking.go#L30)).

RiccardoM:
@HarleyAppleChoi when you have time, please merge the `master` branch here as I've updated it including a lot of unit tests

database/staking.go
```diff
@@ -8,6 +8,7 @@ import (
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 	dbtypes "github.com/forbole/bdjuno/database/types"
 	"github.com/forbole/bdjuno/x/staking/types"
+	bstaking "github.com/forbole/bdjuno/x/staking/types"
```
RiccardoM, line 11: I think we can remove this import as it's the same as the one in the line above, and use `types.` instead of the currenct `bstaking.`

database/staking.go
```diff
@@ -19,40 +20,96 @@ func (db BigDipperDb) SaveStakingPool(pool stakingtypes.Pool, height int64, time
...
+		if err := db.SaveValidatorsData([]types.Validator{validator});err!=nil{
+			return err
+		}
+		return nil
 }
 
 // GetValidatorData returns the validator having the given validator address.
 // If no validator for such address can be found, an error is returned instead.
-func (db BigDipperDb) GetValidatorData(valAddress sdk.ValAddress) (types.Validator, error) {
+func (db BigDipperDb) GetValidatorData(valAddress sdk.ValAddress) (dbtypes.ValidatorData, error) {
```
RiccardoM, line 100: I think that we can revert the change in the returned type, as well as the other `return` statement you have edited here. Since `dbtypes.ValidatorData` implements `types.Validator`, this would be better IMO

database/staking.go
```diff
@@ -97,35 +156,10 @@ func (db BigDipperDb) SaveValidatorsData(validators []types.Validator) error {
...
-	validators := make([]types.Validator, len(rows))
-	for index, row := range rows {
-		validators[index] = row
-	}
-
-	return validators, nil
-}
-
-// _________________________________________________________
-
```
RiccardoM, line 124: Why was this removed entirely? Can we bring it back?

x/staking/types/validator.go
```diff
@@ -88,3 +95,29 @@ type ValidatorDelegations struct {
...
+	MinSelfDelegation int64,Height int64,Timestamp time.Time)(ValidatorCommission){
+	return ValidatorCommission{
+		ValAddress    :ValAddress,
+		Commission 		:Rate,
+		MinSelfDelegation:MinSelfDelegation,
+        Height        :Height,
+        Timestamp     :Timestamp,
+	}
+}
+
```
RiccardoM, line 121: I think we should implement the `Equal` method for this new type. This would be probably useful inside the tests

database/staking.go
HarleyAppleChoi: Duplicate with GetValidatorsData()

database/staking.go
```diff
@@ -48,11 +104,11 @@ func (db BigDipperDb) GetValidatorData(valAddress sdk.ValAddress) (types.Validat
 			 WHERE validator_info.operator_address = $1`
 
 	if err := db.Sqlx.Select(&result, stmt, valAddress.String()); err != nil {
-		return nil, err
+		return dbtypes.ValidatorData{}, err
```
RiccardoM, line 107: I think we should revert this to return `nil` instead.

database/staking.go
```diff
@@ -48,11 +104,11 @@ func (db BigDipperDb) GetValidatorData(valAddress sdk.ValAddress) (types.Validat
...
 	}
 
 	if len(result) == 0 {
-		return nil, fmt.Errorf("no validator with validator address %s could be found", valAddress.String())
+		return dbtypes.ValidatorData{}, fmt.Errorf("no validator with validator address %s could be found", valAddress.String())
```
RiccardoM, line 111: As above

database/staking_test.go
```diff
@@ -174,11 +224,21 @@ func (suite *DbTestSuite) TestBigDipperDb_GetValidatorsData() {
 			"cosmosvalcons1qq92t2l4jz5pt67tmts8ptl4p0jhr6utx5xa8y",
 			"cosmosvaloper1000ya26q2cmh399q4c5aaacd9lmmdqp90kw2jn",
 			"cosmosvalconspub1zcjduepqe93asg05nlnj30ej2pe3r8rkeryyuflhtfw3clqjphxn4j3u27msrr63nk",
+			"",
+			"",
+			"",
+			"",
+			"",
```
RiccardoM, line 231: I think we should specify some values here to be sure they are properly stored and later read

HarleyAppleChoi: Since GetValidatorData means to get back consensus_address, consensus_pubkey,operator_address only, we are not interested in the descriptions. Also there are another method to get back the details, so we dont need to specify this in this test.

RiccardoM: Actually, `GetValidatorData` returns a `types.Validator` object which is defined as follows: 

```golang
// Validator represents a single validator.
// This is defined as an interface so that we can use the SDK types
// as well as database types properly.
type Validator interface {
	GetConsAddr() sdk.ConsAddress
	GetConsPubKey() crypto.PubKey
	GetOperator() sdk.ValAddress	
	GetDescription() staking.Description
}
```

For this reason, we need to test that also the `GetDescription` method returns the proper data.

PR #20, opened 2020-07-09
Apple/self delegation

### LINKED ISSUE #14: Calculate validator self delegation ratio
@kwunyeung inside the requirements we have

> Calculate validator self delegation ratio (staking)

Looking at the `x/staking` API the best one seems to be `/staking/validators?height=446016`: 

http://lcd.morpheus.desmos.network:1317/staking/validators?height=446016

For each validator, what I suppose we should do to compute the self delegation ratio is: 

`self_delefation_ration = tokens / delegator_shares * 100`

Is this correct?

HarleyAppleChoi:
## Description
Implement self delegation and stored into database
A good side effect is that it also stored all delegation for that target validator once someone submit delegation message

## Checklist
Closes #14 
- [x] Wrote unit tests.  


database/staking.go
```diff
@@ -115,32 +115,42 @@ func (db BigDipperDb) SaveValidatorsData(validators []types.Validator) error {
 	validatorQuery := `INSERT INTO validator (consensus_address, consensus_pubkey) VALUES `
 	var validatorParams []interface{}
 
-	validatorInfoQuery := `INSERT INTO validator_info (consensus_address,operator_address,moniker,identity,website,security_contact, details) VALUES`
-
+	validatorInfoQuery := `INSERT INTO validator_info (consensus_address,operator_address,self_delegate_address,moniker,identity,website,security_contact, details) VALUES`
```
RiccardoM, line 118: Since the `self_delegate_address` references the `account` table, we should call `db.SaveUser` or whatever method is that allows us to save that address to make sure that it exists before referencing it during this insertion. Not doing so would later give an error that the account does not exist

database/staking.go
```diff
@@ -439,3 +447,37 @@ func (db BigDipperDb) SaveRedelegations(redelegations []types.Redelegation) erro
...
+	}
+	stmt = stmt[:len(stmt)-1] // Remove the trailing ","
+	stmt += " ON CONFLICT DO NOTHING"
+	_, err := db.Sql.Exec(stmt, delegationShareParam...)
+	if err != nil {
+		return err
+	}
+
+	//insert self delegation in this entry into self_deleation_row
+	//for each time there are someone delegated to that validator, save the validator_self delegation?
```
RiccardoM, line 468: Were you able to answer this question?

database/staking.go
```diff
@@ -439,3 +447,37 @@ func (db BigDipperDb) SaveRedelegations(redelegations []types.Redelegation) erro
...
+	stmt = `INSERT INTO validator_self_delegation (operator_address ,shares,height,timestamp)
+	SELECT validator_delegation_shares.operator_address, shares,height,timestamp
+	FROM validator_delegation_shares INNER JOIN validator_info
+	ON (validator_delegation_shares.delegator_address = validator_info.self_delegate_address
+		AND validator_delegation_shares.operator_address = validator_info.operator_address)
+	WHERE height = $1`
+
+	//Since shares have same timestamp
+	_, err = db.Sql.Exec(stmt, shares[0].Height)
```
RiccardoM, line 477: Why are we using this table? Couldn't we simply use `validator_delegation_shares` and search for `delegator_address == validator.self_delegation_address`?

x/staking/handlers/validators.go
```diff
@@ -16,7 +17,9 @@ func HandleMsgCreateValidator(msg stakingtypes.MsgCreateValidator, db database.B
 		stakingValidator.GetConsAddr(),
 		stakingValidator.GetOperator(),
 		stakingValidator.GetConsPubKey(),
-		stakingValidator.Description))
+		stakingValidator.Description,
+		sdktypes.AccAddress(stakingValidator.GetConsAddr()),
```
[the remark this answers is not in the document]
kwunyeung: The **self delegation address** is the **operator address** in different prefixes. The prefix of **operator address** is `cosmosvaloper` and the prefix of **self delegation address**, which is a wallet address, is `cosmos`. Simply put, the **operator address** is converted from the validator wallet address when an account creates a validator.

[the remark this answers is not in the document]
RiccardoM: Ok, so @HarleyAppleChoi I think we need to change `sdktypes.AccAddress(stakingValidator.GetConsAddr())` to `sdktypes.AccAddress(stakingValidator.GetValAddr())`

database/staking.go
HarleyAppleChoi: SelfDelegationAccQuery := `INSERT INTO account (address) VALUES `...
I am feeling already implement that saving into the account in line 122 in the same function. Do we need to insert it into balance table as well(like db.SaveAccount)?

database/staking.go
HarleyAppleChoi: Oh I just forget to delete the question mark😂

database/staking.go
HarleyAppleChoi: I think this is a more clear and easier way to do more further calculations, like self delegation ratio.
If that is not nesserary we can just delete that.

RiccardoM: @HarleyAppleChoi I think it's fine deleting this, since thanks to the GraphQL APIs that will be built on top of this database, we will have that data in any case. So we can avoid duplicates

database/staking.go
RiccardoM: @HarleyAppleChoi Sorry, it's fine as you've done it here. I simply missed the query since it is declared below. Do you mind switching line 122 with 118 so that we first initialize the `INSERT INTO account` query and then the `INSERT INTO validator_info` query?

PR #22, opened 2020-07-15
Read consensus state

### LINKED ISSUE #12: Read the latest consensus state
Inside the requirements, there's the following: 

> Read the latest consensus state

@kwunyeung can we more details on what you mean by "consensus state"?

RiccardoM:
## Description
This PR implements the reading and storing of the last consensus events (closes #12). 

Clients will be able to subscribe to such changes by subscribing to GraphQL events on the `consensus` table.

## Checklist
- [x] Wrote unit tests.  


x/consensus/operations.go
```diff
@@ -0,0 +1,107 @@
...
+
+// merge takes a list of read-only channels and merges them into a single read-only channel
+func merge(cs ...<-chan tmctypes.ResultEvent) <-chan tmctypes.ResultEvent {
+	out := make(chan tmctypes.ResultEvent)
+
+	var wg sync.WaitGroup
+	wg.Add(len(cs))
+
+	for _, c := range cs {
+		go func(c <-chan tmctypes.ResultEvent) {
```
HarleyAppleChoi, line 93: Why not putting subscribeConsensusEvent here? That will be more efficient

RiccardoM: I think that what you suggest cannot be done. This line is inside the following method: 

```go
// merge takes a list of read-only channels and merges them into a single read-only channel
func merge(cs ...<-chan tmctypes.ResultEvent) <-chan tmctypes.ResultEvent
```

There we are iterating over the list of channels and merging all their emissions into one single channel. By doing so we can than handle easily all the emitted events as the same way as they had been emitted by a single source instead of multiple subscriptions. 

This is done due to the fact that in order to listen to the current consensus, we need to listen to multiple events using the RPC, but for each event we want to listen to, we need to create a subscription. Each subscription returns a read-only channel (a stream of events). When you have multiple channels, the only way to listen to them at the same time and get all the events as soon as they are emitted preserving their order, is to merge such channels into one (at least that's what I found out). 

If you have an example implementation that works as you described I would be more than happy to see it as I didn't find anything different from this

PR #31, opened 2020-07-28
Apple/coingecko

### LINKED ISSUE #27: Token price and market cap
## Feature description 
We need to implement the ability of fetching every 30 seconds the token price and market cap. 

## Implementation proposal
I think that the token denomination can be easily get from the `/staking/parameters` endpoint, taking the `bond_denom` denomination. 

@kwunyeung do you have any preference from where to get the data? Do you know any API that might allow us to fetch it?

HarleyAppleChoi:
## Description
Getting market cap and coin value from coingecko but we should put it on supply module because that depends on the total_supply table.
It is a fork that may rebase to the local copy of coingecko being an independent module.
close #27 
## Checklist
- [x] Wrote unit tests.  


database/supply.go
```diff
@@ -16,3 +16,27 @@ func (db BigDipperDb) SaveSupplyToken(coins sdk.Coins, height int64) error {
 	}
 	return nil
 }
+
+//GetTokenNames get token name from  latest height
```
RiccardoM, line 20:
```suggestion
// GetTokenNames returns the list token names stored inside the supply table
```

database/supply.go
```diff
@@ -16,3 +16,27 @@ func (db BigDipperDb) SaveSupplyToken(coins sdk.Coins, height int64) error {
...
+
+//return the latest height(has 30 second latency because depend on supply module)
+func (db BigDipperDb) GetLatestHeight() (int64, error) {
+	var height []int64
+	query := `select max(height) FROM supply`
+	if err := db.Sqlx.Select(&height, query); err != nil {
+		return -1, err
+	}
+	return height[0], nil
+}
```
RiccardoM, line 42: I think it's better to remove this. If we want to have the latest block height for some reason, it's better to use a `BlockHandler` instead that always has the updated one

schema/07-coingecko.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE token_values(
```
RiccardoM, line 1: I think it's better to rename this table `token_price` instead. This would probably make it easier to understand what it is used for

schema/07-coingecko.sql
```diff
@@ -0,0 +1,6 @@
...
+    denom TEXT,
+    current_price NUMERIC,
```
RiccardoM, line 3: I think we can rename this column to be just `price` since it already has a `height` associated to it

schema/07-coingecko.sql
```diff
@@ -0,0 +1,6 @@
...
+    market_cap NUMERIC,
+    height INT
```
RiccardoM, line 5: I think it's better to use a `date` column instead of a `height` one. Market prices varies based on time, not block height.

x/supply/coinGeckoTypes/coins.go
```diff
@@ -0,0 +1,41 @@
+package coingecko
```
RiccardoM, line 1: I honestly would remove any CoinGecko reference from any file. While it's true that we are currently basing our data fetching on their APIs, it might not be the case in the future. For this reason, I think it's better to rename such references to something more generic. 

Also, I would avoid putting market-related thing inside the `x/supply` folder. Instead, I would create something like `x/pricefeed` which is more generic, putting everything there.

x/supply/coinGeckoTypes/coins.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+}
+
+// Coins is an array of Coin
+type Coins []Coin
+
+//____________________________________________________
+
+// Market represent some of the attributes in coingecko api's coins/market
+type Market struct {
```
RiccardoM, line 25: I would probably rename this to be `MarketTicker` or something else more specific. Right now it doesn't allow to understand exactly what it represents

x/supply/operations/price.go
```diff
@@ -0,0 +1,76 @@
...
+		if err != nil {
+			return err
+		}
+		// Store the signing infos into the database
+		err = db.SaveSupplyToken(s,
+			height,
+		)
+		if err != nil {
+			return err
+		} */
```
RiccardoM, line 74: Can we remove this?

x/supply/operations/price.go
```diff
@@ -0,0 +1,76 @@
...
+	//query
+	var markets api.Markets
+	query := fmt.Sprintf("/coins/markets?vs_currency=usd&ids=%s", ids)
+	if err = utils.QueryCoinGecko(query, &markets); err != nil {
+		return err
+	}
+
+	if err = db.SaveTokensPrice(markets, height); err != nil {
+		return err
+	}
```
RiccardoM, line 60: I think it's better to easier to return this result: 

```golang
return db.SaveTokensPrice(markets, height)
```

x/utils/httpQueries.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// QueryCoinGecko can query endpoint from coingecko
+func QueryCoinGecko(endpoint string, ptr interface{}) error {
+	url := "https://api.coingecko.com/api/v3"
+	if err := queryFromAPI(url, endpoint, ptr); err != nil {
+		return err
+	}
+	return nil
+}
```
RiccardoM, line 40: I don't think this should be put inside the `x/utils` folder since it will only be used by the price-related methods. Instead, it should be placed inside the module that uses it. We should place only generic methods here that can be used by a variety of different modules.

x/utils/httpQueries.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+)
+
+//queryFromAPI is generic function that can query every api and endpoint
+func queryFromAPI(url string, endpoint string, ptr interface{}) error {
```
RiccardoM, line 12: I think that having the `url` and `endpoint` separated is not the best way to ask users for such parameters. Instead, it would be much easier for them to speify a single `endpoint` parameter that already contains the `url` part itself.

database/types/coingecko.go
```diff
@@ -0,0 +1,27 @@
...
+func NewTokenPriceRow(denom string, currentprice float64, marketcap int64, height int64) TokenPriceRow {
+	return TokenPriceRow{
+		Denom:        denom,
+		Currentprice: currentprice,
+		Marketcap:    marketcap,
+		Height:       height,
+	}
+}
+
+// Equals return true if one NewTokenPriceRow representing the same row as the original one
```
RiccardoM, line 21:
```suggestion
// Equals return true if u and v represent the same row
```

database/supply.go
HarleyAppleChoi: I actually don't know if saving block height or timestamp will be better for this case. Maybe just storing timestamps is better?

RiccardoM: @HarleyAppleChoi Yeah, I think timestamp would be better in this case

PR #32, opened 2020-07-29
Apple/coingecko2

### LINKED ISSUE #27: Token price and market cap

HarleyAppleChoi:
## Description

A fork of #31 that makes a stand alone module of x/pricefeed
close #31 
close #27 
## Checklist
- [x] Wrote unit tests.  


database/pricefeed.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	api "github.com/forbole/bdjuno/x/pricefeed/apiTypes"
+)
+
+// SaveTokensPrice allows to save for the given height the given total amount of coins
+func (db BigDipperDb) SaveTokensPrice(pricefeeds api.Pricefeeds, timestamp time.Time) error {
```
RiccardoM, line 11: Since we're storing only one market ticker (price, market cap, etc) at a time, I think it would be better here to only accept a single `types.MarketTicker` instead of `Pricefeeds` (which represent a list of them). This would be more correct overall IMO and it would also allow you to write simpler code in the body of the method

PR #35, opened 2020-08-17
Apple/query enable modules

### LINKED ISSUE #4: Ability for querying the enabled modules
With the modular structure, I would like to have a way to query the enabled modules in the `BDJuno`. The idea is that once the modules is enabled in `BDJuno`, a `GraphQL` query will return the list of modules enabled, so that the frontend application will know which corresponding components should be enabled. We are going to develop the frontend following the same modular structure.

HarleyAppleChoi:
## Description
Close #4 
## Checklist
- [x] Wrote unit tests.  


RiccardoM (commented):
@HarleyAppleChoi I've taken a look at the implementation, and I got some comments on it. 

### Modules storing
First of all, I don't understand why you decided to store the enabled modules inside the database in that way. I personally don't think that having a single row with multiple columns is the best thing. 

Instead, what we could do is have a single column (acting as a `PRIMARY KEY`) that contains the module names. This way instead of having 

| Staking | Auth | Supply | Distribution | Pricefeed | Bank |  Consensus | Mint | Timestamp |
| :----: | :----: | :----: | :----: | :----: | :----: | :----: | :----: | :----: |
| true | false | true | true | true | false | false | true | 2020-08-19T12:00Z |

We could have something like 

| Module name | 
| :-------: | 
| staking | 
| supply | 
| distribution | 
| pricefeed | 
| mint | 

I think that saving only the enabled modules this way will make everything easier to manage. If a new module is enabled, we simply add a row. If it is disabled, we remove that row. Having to add/remove a column would require a database + GraphQL migration each time, which is quite complex to perform. 

Also, this way we could have the GraphQL client observing the enabled modules using GQL subscription, and changing it's UI dynamically. This is not possible using the current implementation. 

### Modules registration
Currently in order to register the different modules you have created an `AdditionalOperation` and registered the modules by hand. While this works, I don't think this is the best way to go. 

BDJuno will act as a `daemon`. This means that it will be started once and then it will continue to run until it is blocked. For this reason, setting the different `true/false` values by hand is not the best way to go. Also, if we consider what I've written above it makes even less sense. 

I think that the best way to approach the modules registration is to define a new interface as follows: 

```go
type Module interface { 
  Name() string
  BlockHandlers() []juno.BlockHandlers
  TxHandlers() []juno.TxHandlers
  MsgHandlers() []juno.TxHandlers
  AdditionalOperations() []parse.AdditionalOperation
  PeriodicOperations() []PeriodicOperations
}
```

This module interface would then need to be implemented inside all the different `x/` modules. Here's an example (`x/staking/module.go`): 

```go
type StakingModule struct {}

func (m StakingModule) Name() {
  return "staking" 
} 

func (m StakingModule) BlockHandlers() []juno.BlockHandler {
  return juno.BlockHandler[]{ BlockHandler }
} 

// and so on...
```

Once that is being done, what we can do is we can define a method as follows: 

```go
func RegisterModules(modules []Module) {
  // store inside the DB the enabled module
  // register each block handler
  // register each tx handler
  // and so on...
}
```

Finally, we simply have to call this method from the `main` function: 

```go
  RegisterModules(staking.Module{}, distr.Module{}, ...)
```

This would take care of everything and would allow us in the future to load the modules dynamically based on a configuration JSON or other things. 

### Conclusions 
I would revisit what you have done following my tips. Personally, what I would like you to do is the following: 

1. Try implementing the database as I've suggested. 
2. Once implemented, ask me a review of the one. 
3. When the review is approved, then proceed to the implementation of the new registration methodology.

RiccardoM (approved):
Looks good @HarleyAppleChoi, you can go ahead with the new registration implementation :100:

RiccardoM (approved):
Looks fantastic :rocket:  Great job @HarleyAppleChoi :100:

PR #45, opened 2020-08-28
Updates and fixes

RiccardoM:
## Description
- Updated Cosmos to v0.39.1
- Merged the typos fix branch
- Fixed the inflation not working (closes #36)

## Checklist
- [ ] Targeted PR against correct branch.
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Wrote unit tests.  
- [ ] Re-reviewed `Files changed` in the Github PR explorer.

PR #46, opened 2020-08-28
Apple/fix typo

### LINKED ISSUE #37: Data errors in various big.Int datatype
## Bug description
If you are using big.Int.int64 that is not the real value, especially on the validator_delegation module.

## Expected behavior
Expected much smaller value than 80000000000000

HarleyAppleChoi:
## Description

It turns out the problem with dbcoin that not handling big.Int good enough.
close #37 
## Checklist
- [ ] Targeted PR against correct branch.
- [ ] Wrote unit tests.  


HarleyAppleChoi:
Hm that maybe ok with the original one actually
just leave that and check later🧐

RiccardoM:
Included in #45

PR #48, opened 2020-09-02
Apple/last update validator

### LINKED ISSUE #47: Save the time when validator update description
## Feature description 
Save time and block height when the validator updates his description so that we can retrieve the description that the validator has made.

HarleyAppleChoi:
## Description
close #47
That is better to have a seperate table for validator description so that historical data can be retrieve
## Checklist
- [x] Wrote unit tests.  


RiccardoM (approved):
LGTM :rocket:

schema/03-staking.sql
```diff
@@ -82,3 +77,15 @@ CREATE TABLE validator_voting_power
     height            BIGINT NOT NULL,
     PRIMARY KEY (consensus_address, height)
 );
+
+CREATE TABLE validator_description
+(
+    operator_address      TEXT NOT NULL,
```
RiccardoM, line 83: @HarleyAppleChoi Can we add `references validator_info(operator_address)` here? This would allow Hasura to automatically create a relationship and include the `description` field inside the `validator_info` object. 

```suggestion
    operator_address      TEXT NOT NULL REFERENCES validator_info(operator_address),
```

schema/03-staking.sql
```diff
@@ -82,3 +77,15 @@ CREATE TABLE validator_voting_power
...
+    moniker               TEXT ,
+    identity              TEXT,
+    website               TEXT,
+    security_contact      TEXT,
+    details               TEXT,
+    height                BIGINT,
+    timestamp             TIMESTAMP
```
RiccardoM, line 90: What about switching from `TIMESTAMP` to `TIMESTAMP WITHOUT TIME ZONE`? I would also add the `NOT NULL` clause here. 

```suggestion
    timestamp             TIMESTAMP WITHOUT TIME ZONE NOT NULL
```

RiccardoM (approved):
:100:

PR #162, opened 2021-08-04
Dockerise Bdjuno & create docker-compose for running Bdjuno and hasura

maiquanghiep:
## Description
Dockerise Bdjuno

## Checklist
- [ ] Wrote unit tests.  

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,10 @@
-FROM golang:1.16 AS builder
-WORKDIR /go/src/github.com/forbole/big-dipper
+FROM golang:1.16-alpine AS builder
+RUN apk update && apk add --no-cache make git
+WORKDIR /go/src/github.com/forbole/bdjuno
 COPY . ./
-RUN CGO_ENABLED=0 go build -o /build/bdjuno ./cmd/bdjuno
+RUN make build
 
 FROM alpine:latest
-WORKDIR /root/
-COPY --from=0 /build/bdjuno ./
-RUN ./bdjuno init
-ENTRYPOINT ["./bdjuno"]
-CMD ["parse", "--home", "/root/.bdjuno"]
+WORKDIR /bdjuno
+COPY --from=builder /go/src/github.com/forbole/bdjuno/build/bdjuno /usr/bin/bdjuno
+CMD [ "bdjuno" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM golang:1.16 AS builder
+WORKDIR /go/src/github.com/forbole/big-dipper
+COPY . ./
+RUN CGO_ENABLED=0 go build -o /build/bdjuno ./cmd/bdjuno
+
+FROM alpine:latest
+WORKDIR /root/
+COPY --from=0 /build/bdjuno ./
+RUN ./bdjuno init
+ENTRYPOINT ["./bdjuno"]
+CMD ["parse", "--home", "/root/.bdjuno"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,23 @@
+version: '3.6'
+services:
+  hasura:
+    image: hasura/graphql-engine:v2.0.4
+    ports:
+    - "8080:8080"
+    restart: always
+    environment:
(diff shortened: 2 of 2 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+      HASURA_GRAPHQL_DATABASE_URL: postgres://${POSTGRES_USER}:${POSTGRES_PASSWORD}@postgres:5432/${POSTGRES_DB}
+      ## enable the console served by server
+      HASURA_GRAPHQL_ENABLE_CONSOLE: "true" # set to "false" to disable console
+      ## enable debugging mode. It is recommended to disable this in production
+      HASURA_GRAPHQL_DEV_MODE: "true"
+      HASURA_GRAPHQL_ENABLED_LOG_TYPES: startup, http-log, webhook-log, websocket-log, query-log
+      ## uncomment next line to set an admin secret
+      # HASURA_GRAPHQL_ADMIN_SECRET: myadminsecretkey
+  bdjuno:
+    image: forbole/bdjuno:latest
```
maiquanghiep: Didn't push to docker hub yet

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.16 AS builder
+WORKDIR /go/src/github.com/forbole/big-dipper
+COPY . ./
+RUN CGO_ENABLED=0 go build -o /build/bdjuno ./cmd/bdjuno
```
rmontagnin: Instead of running this command, we should use `make build` instead and let the Makefile to take care of properly building the command. This will ensure that the binaries are always built using the same method. 

```suggestion
RUN make build
```

Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.16 AS builder
+WORKDIR /go/src/github.com/forbole/big-dipper
```
rmontagnin: I think we can change the `WORKDIR` to be something related to BDJuno instead. Maybe its GitHub repository path: 

```suggestion
WORKDIR /go/src/github.com/forbole/bdjuno
```

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+COPY . ./
+RUN CGO_ENABLED=0 go build -o /build/bdjuno ./cmd/bdjuno
+
+FROM alpine:latest
+WORKDIR /root/
```
rmontagnin: I think it's always better to move out of the `/root` folder as a workdir, since it might have special rights being owned only by the `root` user:

```suggestion
WORKDIR /bdjuno
```

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+COPY --from=0 /build/bdjuno ./
```
rmontagnin: I would suggest using named containers instead of int indexes here. Also, we can simply move the `bdjuno` binary directly into the `/usr/bin` folder to make it immediately executable and known to the system as a whole: 

```suggestion
COPY --from=builder /go/src/github.com/forbole/bdjuno/build/bdjuno /usr/bin/bdjuno
```

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+RUN ./bdjuno init
+ENTRYPOINT ["./bdjuno"]
+CMD ["parse", "--home", "/root/.bdjuno"]
```
rmontagnin: Instead of taking care about initializing the configuration, I think it would be much better to leave the ability to run such commands to the external user. We can simply set the `CMD` to be `bdjuno` so that the user can use the container to run all the other commands (`init`, `parse`, etc) by their own:

```suggestion
CMD [ "bdjuno" ]
```

Dockerfile
maiquanghiep: Do you have any recommendations for the base image in stage 2? Without CGO_ENABLED=0, I get the error: "standard_init_linux.go:228: exec user process caused: no such file or directory" in alpine.

rmontagnin: I think this is due to the fact that you are using `golang:1.16` to build the binary, and then trying to run such binary on an alpine image. I would suggest using `golang:1.16-alpine` instead so that it links to the proper OS libraries

rmontagnin (approved):

PR #199, opened 2021-09-20
feat: allow to set delegators rewards query frequency

MonikaCat:
## Checklist
- [ ] Wrote unit tests.  

### DIFF
--- a/database/distribution.go
+++ b/database/distribution.go
@@ -198,15 +198,16 @@ func (db *Db) DeleteDelegatorRewardsAmount(delegatorAddr string, height int64) e
 	return nil
 }
 
-// GetDelegatorRewards returns all delegation reward data stored in db
-func (db *Db) GetDelegatorRewards() ([]*dbtypes.DelegationRewardRow, error) {
-	stmt := `SELECT * FROM delegation_reward`
+// HasDelegatorRewards checks if data is present in db and returns boolean value
+func (db *Db) HasDelegatorRewards() (bool, error) {
+	stmt := `SELECT COUNT(*) FROM delegation_reward`
 
-	var rows []*dbtypes.DelegationRewardRow
-	err := db.Sqlx.Select(&rows, stmt)
+	var count int
+	err := db.Sql.QueryRow(stmt).Scan(&count)
 	if err != nil {
-		return nil, err
+		return false, fmt.Errorf("error while getting delegation reward: %s", err)
+
 	}
 
-	return rows, nil
+	return count > 0, nil
 }
--- a/database/distribution.go
+++ b/database/distribution.go
@@ -210,16 +210,3 @@ func (db *Db) GetDelegatorRewards() ([]*dbtypes.DelegationRewardRow, error) {
 
 	return rows, nil
 }
-
-// GetDelegatorRewards returns all delegation reward data stored in db
-func (db *Db) GetDelegatorRewards() ([]*dbtypes.DelegationRewardRow, error) {
-	stmt := `SELECT * FROM delegation_reward`
-
-	var rows []*dbtypes.DelegationRewardRow
-	err := db.Sqlx.Select(&rows, stmt)
-	if err != nil {
-		return nil, err
-	}
-
-	return rows, nil
-}
--- a/database/distribution.go
+++ b/database/distribution.go
@@ -198,7 +198,7 @@ func (db *Db) DeleteDelegatorRewardsAmount(delegatorAddr string, height int64) e
 	return nil
 }
 
-// HasDelegatorRewards checks if data is present in db and returns boolean value
+// HasDelegatorRewards checks if the database contains any delegation reward
 func (db *Db) HasDelegatorRewards() (bool, error) {
 	stmt := `SELECT COUNT(*) FROM delegation_reward`
 
--- a/database/distribution.go
+++ b/database/distribution.go
@@ -197,3 +197,16 @@ func (db *Db) DeleteDelegatorRewardsAmount(delegatorAddr string, height int64) e
 
 	return nil
 }
+
+// GetDelegatorRewards returns all delegation reward data stored in db
+func (db *Db) GetDelegatorRewards() ([]*dbtypes.DelegationRewardRow, error) {
+	stmt := `SELECT * FROM delegation_reward`
+
+	var rows []*dbtypes.DelegationRewardRow
+	err := db.Sqlx.Select(&rows, stmt)
+	if err != nil {
+		return nil, err
+	}
+
+	return rows, nil
+}
--- a/database/distribution.go
+++ b/database/distribution.go
@@ -177,3 +177,16 @@ func (db *Db) DeleteDelegatorRewardsAmount(delegatorAddr string, height int64) e
 	_, err := db.Sql.Exec(stmt, delegatorAddr, height)
 	return err
 }
+
+// GetDelegatorRewards returns all delegation reward data stored in db
+func (db *Db) GetDelegatorRewards() ([]*dbtypes.DelegationRewardRow, error) {
+	stmt := `SELECT * FROM delegation_reward`
+
+	var rows []*dbtypes.DelegationRewardRow
+	err := db.Sqlx.Select(&rows, stmt)
+	if err != nil {
+		return nil, err
+	}
+
+	return rows, nil
+}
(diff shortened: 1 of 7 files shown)


modules/distribution/utils/delegator_rewards.go
```diff
@@ -13,8 +13,27 @@ import (
...
+		// run once when chain starts
+		go updateDelegatorsRewards(height, client, db)
+	}
+
+	if len(rewards) > 0 {
+		// run every 100 blocks interval
+		if height%BlockInterval == 0 {
+			go updateDelegatorsRewards(height, client, db)
+		}
+	}
```
rmontagnin: I think this can be simplified as follows: 

```suggestion
	if len(rewards) == 0 || height%BlockInterval == 0 {
		go updateDelegatorsRewards(height, client, db)
	}
```

Also, I think we should move the `BlockInterval` value somewhere else. Maybe we can set it inside the module's constructor method so that in other chains (e.g. `Osmosis`) we can get the value from the database when first creating the module?

rmontagnin: Maybe even a better alternative is to have a custom BDJuno params table inside the database that holds this values. This way we can even play with them to find the best value without pushing out new versions of BDJuno in the future

MonikaCat: BlockInterval value will always be hardcoded, so how about adding the value to bdjuno config file and saving it to distribution_params table from there?

rmontagnin: @MonikaCat Why is it hardcoded? I mean, we might want to update the delegations on Cosmos every 30 blocks and on Terra every 100. Also, on epoch-based chains like Osmosis the epoch interval will surely be present inside on-chain params

MonikaCat: I tried to query epoch length value for osmosis chain but I can't seem to find it

MonikaCat: Can you point me to the correct function for that?

rmontagnin: @MonikaCat I think you can get it by querying the [`mint` params](https://github.com/osmosis-labs/osmosis/blob/cf3246aaad8c2269225b430a1c3dd776a172ba63/x/mint/spec/04_params.md). This contains a `epoch_identifier` that can be used to query the details of such epoch from the [`epoch` module](https://github.com/osmosis-labs/osmosis/blob/cf3246aaad8c2269225b430a1c3dd776a172ba63/x/epochs/spec/02_state.md).

MonikaCat: Cool! I updated osmosis PR #198 with querying ```epoch_identifier``` value instead of hard coding ```BlockInterval``` value.

MonikaCat: The base branch however will need to have this value hard coded because, as you mentioned already, we might want to adjust it for each chain individually.

rmontagnin: @MonikaCat I think it would be way better to extend the current configuration allowing us to set such value from there instead of hard coding the values. Something like this would be perfect in my opinion: 

```toml
[distribution]
rewards_frequency = 100
```

This way, in the future we an also add different frequencies for other queries and modules as well. What do you think?

MonikaCat: Yup I agree! It would also help users to setup bdjuno config in one place, without the need to modify the source code. Alright I will update the code with the above :)

MonikaCat: I added the rewards_frequency field in config.toml file and updated the source code, what do you think?

database/distribution.go
```diff
@@ -197,3 +197,16 @@ func (db *Db) DeleteDelegatorRewardsAmount(delegatorAddr string, height int64) e
 
 	return nil
 }
+
+// GetDelegatorRewards returns all delegation reward data stored in db
+func (db *Db) GetDelegatorRewards() ([]*dbtypes.DelegationRewardRow, error) {
```
rmontagnin: Since we are using this only to check if there are some delegations, I think it would be much more efficient to have a method such as 

```go
func (db *Db) HasDelegatorRewards() bool
```

Which simply returns `true` if there are some rewards, and `false` otherwise. This can then be implemented with a more efficient query as 

```sql
SELECT COUNT(*) FROM delegation_reward
```

And then simply return `count > 0`

modules/distribution/handle_block.go
```diff
@@ -12,17 +12,18 @@ import (
 
 	"github.com/forbole/bdjuno/database"
 	distrutils "github.com/forbole/bdjuno/modules/distribution/utils"
+	"github.com/forbole/bdjuno/types/config"
 )
 
 // HandleBlock represents a method that is called each time a new block is created
-func HandleBlock(block *tmctypes.ResultBlock, client distrtypes.QueryClient, db *database.Db) error {
+func HandleBlock(block *tmctypes.ResultBlock, client distrtypes.QueryClient, db *database.Db, cfg *config.Config) error {
```
rmontagnin: Can we have the `cfg` parameter as the first parameter in order to keep it consistent with other methods as well? I tend to always use the same order of parameters: 

1. configuration objects
2. block/tx/message
3. query clients
4. codecs
5. database

The same also applied elsewhere

modules/distribution/utils/delegator_rewards.go
```diff
@@ -7,14 +7,26 @@ import (
...
-func UpdateDelegatorsRewardsAmounts(height int64, client distrtypes.QueryClient, db *database.Db) {
+func UpdateDelegatorsRewardsAmounts(height int64, client distrtypes.QueryClient, db *database.Db, cfg *config.Config) {
+	rewards, _ := db.GetDelegatorRewards()
+	interval := cfg.GetRewardsFrequencyConfig().GetRewardsFrequency()
+
+	if interval > 0 {
+		if len(rewards) == 0 || height%interval == 0 {
+			go updateDelegatorsRewards(height, client, db)
+		}
+	}
```
rmontagnin: Instead of putting the longer term inside the `if` statement, it's more efficient to return immediately if the condition is not satified: 

```suggestion
	interval := cfg.GetRewardsFrequencyConfig().GetRewardsFrequency()
	if interval == 0 {
		log.Debug().Str("module", "distribution").Msg("delegator rewards refresh interval set to 0. Skipping refresh")
		return
	}
	
	rewards, _ := db.GetDelegatorRewards()
	if len(rewards) == 0 || height%interval == 0 {
		go updateDelegatorsRewards(height, client, db)
	}
```

types/config/config.go
```diff
@@ -102,3 +113,13 @@ type PricefeedConfig struct {
 func (p *PricefeedConfig) GetTokens() []types.Token {
 	return p.Tokens
 }
+
+// RewardsFrequencyConfig contains the configuration about rewards frequency in distribution module
+type RewardsFrequencyConfig struct {
```
rmontagnin: It's probably better to rename this to something like `DistributionConfig` in order to keep it consistent with what it contains (the configuration of the `distribution` module). We might want to use it also for other refresh times in the future as well.

types/config/config.go
```diff
@@ -12,18 +12,21 @@ var _ juno.Config = &Config{}
 // Config contains the data about the BDJuno configuration
 type Config struct {
 	juno.Config
-	PricefeedConfig *PricefeedConfig
+	PricefeedConfig        *PricefeedConfig
+	RewardsFrequencyConfig *RewardsFrequencyConfig
```
rmontagnin: If the struct ends up being renamed, this should be changed too

modules/distribution/handle_block.go
MonikaCat: Ok 👍🏻

modules/distribution/utils/delegator_rewards.go
MonikaCat: Updated

database/distribution.go
```diff
@@ -198,15 +198,16 @@ func (db *Db) DeleteDelegatorRewardsAmount(delegatorAddr string, height int64) e
 	return nil
 }
 
-// GetDelegatorRewards returns all delegation reward data stored in db
-func (db *Db) GetDelegatorRewards() ([]*dbtypes.DelegationRewardRow, error) {
-	stmt := `SELECT * FROM delegation_reward`
+// HasDelegatorRewards checks if data is present in db and returns boolean value
```
rmontagnin:
```suggestion
// HasDelegatorRewards checks if the database contains any delegation reward
```

rmontagnin (approved):

PR #207, opened 2021-09-24
feat: add support for e-Money Gas Prices

### LINKED ISSUE #202: [e-Money] Get minimum gas prices
## Feature description 

Log the minimum gas prices from x/authority

```
emd q authority gas-prices
```

```
{
  "min_gas_prices": [
    {
      "denom": "echf",
      "amount": "0.530000000000000000"
    },
    {
      "denom": "edkk",
      "amount": "3.700000000000000000"
    },
    {
      "denom": "eeur",
      "amount": "0.500000000000000000"
    },
    {
      "denom": "enok",
      "amount": "6.100000000000000000"
    },
    {
      "denom": "esek",
      "amount": "5.500000000000000000"
    },
    {
      "denom": "ungm",
      "amount": "1.000000000000000000"
    }
  ]
}
```

huichiaotsou:
## Description
Fixes #202 

## Checklist
- [x] Wrote unit tests.  


MonikaCat:
Tested on my local ACK

huichiaotsou:
Hello @RiccardoM , I have completed the e-Money min gas prices module with hasura metadata and unit tests, but I'm not sure why there's a linter error :-/

rmontagnin (approved):
utACK

rmontagnin:
Can I get a new confirmation from @MonikaCat that this works please?

MonikaCat:
> Can I get a new confirmation from @MonikaCat that this works please?

I tested it on my local, it works. However, should we add periodic operation to query the gas prices once a day or so?

rmontagnin:
> I tested it on my local, it works. However, should we add periodic operation to query the gas prices once a day or so?

To be honest, I have no idea how this module works and what it represents. I see that @huichiaotsou added a message handler for gas prices. Is that necessary, or should it be replace with a periodic query instead?

huichiaotsou:
> > I tested it on my local, it works. However, should we add periodic operation to query the gas prices once a day or so?
> 
> To be honest, I have no idea how this module works and what it represents. I see that @huichiaotsou added a message handler for gas prices. Is that necessary, or should it be replace with a periodic query instead?

So e-Money has an authority module in which they set the gas prices : https://github.com/e-money/em-ledger/blob/master/x/authority/handler.go#L31

which is why I used message handler to fetch the msg SetGasPrices. I didn't set up periodic ops as I thought it wouldn't be updated so frequently ? 😆

MonikaCat:
Yes this approach will work if we are setting up bdjuno from genesis time. Bdjuno will process the genesis file and save the gas values in db. However, if we update bdjuno that is currently running with it, we will need to rely on/wait for the gas prices to be updated before we can store the values in db. Do you know how often they are updated? 
Also, I can't seem to find any transactions with this msg in hasura ```/em.authority.v1.MsgSetGasPrices```? Is this the correct msg type?

modules/authority/handle_periodic_operations.go
```diff
@@ -0,0 +1,54 @@
...
+	if _, err := scheduler.Every(1).Day().At("00:00").StartImmediately().Do(func() {
+		utils.WatchMethod(func() error { return updateMinGasPrices(authorityclient, db) })
+	}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// updateInflation fetches from the REST APIs the latest value for the
```
MonikaCat: I guess this should be ```updateMinGasPrices```

huichiaotsou: thanks~ it's updated :D

MonikaCat:
Tested on my local with periodic ops added, ACK

modules/authority/module.go
```diff
@@ -0,0 +1,62 @@
...
+	return HandleGenesis(genesisDoc, appState, m.encodingConfig.Marshaler, m.db)
+}
+
+// RegisterPeriodicOperations implements modules.PeriodicOperationsModule
+func (m *Module) RegisterPeriodicOperations(scheduler *gocron.Scheduler) error {
+	return RegisterPeriodicOps(scheduler, m.authorityClient, m.db)
+}
+
+// HandleMsg implements modules.MessageModule
+func (m *Module) HandleMsg(_ int, msg sdk.Msg, tx *types.Tx) error {
```
rmontagnin: Should we delete this since we are already updating them periodically?

huichiaotsou: hello @RiccardoM I have removed the handle msg :D

rmontagnin (approved):

PR #214, opened 2021-10-01
[e-Money] fix: gas prices not being updated periodically

### LINKED ISSUE #202: [e-Money] Get minimum gas prices

### LINKED ISSUE #213: [e-Money] Gas prices periodic ops not working without genesis parsed
## Bug description

the periodic operations handler for gas prices is currently not working because it only handles UPDATE the table on gas_prices & height columns, whereas the genesis being not parsed, the initial row is not there so the UPDATE will fail. 

## Expected behavior

Checked with Wingman that they won't need the authority_key, planning to remove the column from the table, leaving only gas_prices and height

huichiaotsou:
## Description
Fixes #213, closes #202 

## Checklist
- [x] Wrote unit tests.  

### DIFF
--- a/database/emoney_authority.go
+++ b/database/emoney_authority.go
@@ -3,7 +3,6 @@ package database
 import (
 	"fmt"
 
-	sdk "github.com/cosmos/cosmos-sdk/types"
 	dbtypes "github.com/forbole/bdjuno/database/types"
 	"github.com/forbole/bdjuno/types"
 	"github.com/lib/pq"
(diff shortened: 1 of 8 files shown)


huichiaotsou:
hello @MonikaCat, this morning I realized that my periodic operations handler for gas prices would not work because it [only handles UPDATE table](https://github.com/forbole/bdjuno/blob/chains/emoney/mainnet/database/emoney_authority.go#L32)(**authority_key** can only be obtained from the genesis, so the periodic operations could only use UPDATE). However there is no row to be UPDATEd because genesis was not parsed... the UPDATE then will fall in error. 

So I discussed with @ryuash and she confirmed that they won't need the **authority_key** so I removed it from the type struct as well as from the table. Now handle_genesis and handle_periodic_operations both use INSERT INTO ON CONFLICT so it will prevent the error.

rmontagnin:
@huichiaotsou Can we leave the `authority_key` field and still use the `one_row_id`? I think other people might want to have that data (i.e. if they are customizing BigDipper). 

Also, will this close #202 as well? If so, can you add it to the description please?

huichiaotsou:
> @huichiaotsou Can we leave the `authority_key` field and still use the `one_row_id`? I think other people might want to have that data (i.e. if they are customizing BigDipper).
> 
Sure, I will add this col in the table and still include insert of this data from handle_genesis 
as for PeriodicOperations I will then leave authority_key un-updated.

> Also, will this close #202 as well? If so, can you add it to the description please?
>
OK added!

rmontagnin:
> Sure, I will add this col in the table and still include insert of this data from handle_genesis as for PeriodicOperations I will then leave authority_key un-updated.

Hmm not sure I quite understand then. Isn't this key retrieved from the querier that gets the gas prices? If not then we can leave it out. I thought that is was returned by the querier cause it would have made sense to indicate who was the last one updating the prices

huichiaotsou:
> > Sure, I will add this col in the table and still include insert of this data from handle_genesis as for PeriodicOperations I will then leave authority_key un-updated.
> 
> Hmm not sure I quite understand then. Isn't this key retrieved from the querier that gets the gas prices? If not then we can leave it out. I thought that is was returned by the querier cause it would have made sense to indicate who was the last one updating the prices

😆  no, the query only returns gas prices but no authority key

rmontagnin:
> laughing no, the query only returns gas prices but no authority key

Ok then we can leave it as it is. If the work has done, you can mark this PR as "Ready for review" :smile:

huichiaotsou:
> > laughing no, the query only returns gas prices but no authority key
> 
> Ok then we can leave it as it is. If the work has done, you can mark this PR as "Ready for review" 😄

perfect ! just switched

PR #222, opened 2021-10-07
x/ditr handleMsg add go routines for refreshing rewards and commission

### LINKED ISSUE #215: Handle MsgWithdrawDelegatorReward and MsgWithdrawValidatorCommission
## Feature description 

Because we are no longer getting rewards at every block when a user MsgWithdrawDelegatorReward/ MsgWithdrawValidatorCommission the rewards are not correctly updated right away. We see a lag in the ui that can be very misleading.
## Implementation proposal

If a user runs MsgWithdrawDelegatorReward we should update the rewards

huichiaotsou:
## Description
fixes #215 

## Checklist
- [ ] Wrote unit tests.  

### DIFF
--- a/modules/distribution/handle_block.go
+++ b/modules/distribution/handle_block.go
@@ -17,25 +17,10 @@ func (m *Module) HandleBlock(b *tmctypes.ResultBlock, _ []*juno.Tx, _ *tmctypes.
 	// Update the validator commissions
 	go m.updateValidatorsCommissionAmounts(b.Block.Height)
 
-	// Get interval cfg which decides when to execute "go m.refreshDelegatorsRewardsAmounts(b.Block.Height)"
-	interval := m.cfg.RewardsFrequency
-	if interval == 0 {
-		log.Debug().Str("module", "distribution").Msg("delegator rewards refresh interval set to 0. Skipping refresh")
-		return nil
+	// Update the delegators commissions amounts upon reaching interval or no rewards saved yet
+	if m.shouldUpdateDelegatorRewardsAmounts(b.Block.Height) {
+		go m.refreshDelegatorsRewardsAmounts(b.Block.Height)
 	}
-
-	hasRewards, err := m.db.HasDelegatorRewards()
-	if err != nil {
-		log.Error().Str("module", "distribution").Err(err).Int64("height", b.Block.Height).
-			Msg("error while checking delegators reward")
-	}
-
-	if hasRewards && b.Block.Height%interval != 0 {
-		return nil
-	}
-
-	// Update the delegators commissions amounts
-	go m.refreshDelegatorsRewardsAmounts(b.Block.Height)
 
 	return nil
 }
--- a/modules/distribution/handle_block.go
+++ b/modules/distribution/handle_block.go
@@ -17,8 +17,25 @@ func (m *Module) HandleBlock(b *tmctypes.ResultBlock, _ []*juno.Tx, _ *tmctypes.
 	// Update the validator commissions
 	go m.updateValidatorsCommissionAmounts(b.Block.Height)
 
+	//get interval cfg and verify if func refreshDelegatorsRewardsAmounts() should be executed
+	interval := m.cfg.RewardsFrequency
+	if interval == 0 {
+		log.Debug().Str("module", "distribution").Msg("delegator rewards refresh interval set to 0. Skipping refresh")
+		return nil
+	}
+
+	hasRewards, err := m.db.HasDelegatorRewards()
+	if err != nil {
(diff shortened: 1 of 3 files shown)


modules/distribution/utils_delegator_rewards.go
```diff
@@ -10,21 +10,23 @@ import (
...
-	hasRewards, err := m.db.HasDelegatorRewards()
-	if err != nil {
-		log.Error().Str("module", "distribution").Err(err).Int64("height", height).
-			Msg("error while checking delegators reward")
-	}
-
-	if hasRewards && height%interval != 0 {
-		return
+func (m *Module) refreshDelegatorsRewardsAmounts(height int64, checkInterval bool) {
+	if checkInterval {
```
rmontagnin: I don't like much this approach of having a boolean field just to tell whether or not to check the interval before updating the amounts. Instead, I think the whole interval check can be safely moved inside the `HandleBlock` method since it's used only there. Then this method can become more easily readable and can be safely called from the messages handlers as well

huichiaotsou: Hi @RiccardoM , sure I just modified, is it like this?

modules/distribution/handle_block.go
```diff
@@ -17,6 +17,23 @@ func (m *Module) HandleBlock(b *tmctypes.ResultBlock, _ []*juno.Tx, _ *tmctypes.
...
+	hasRewards, err := m.db.HasDelegatorRewards()
+	if err != nil {
+		log.Error().Str("module", "distribution").Err(err).Int64("height", b.Block.Height).
+			Msg("error while checking delegators reward")
+	}
+
+	if hasRewards && b.Block.Height%interval != 0 {
+		return nil
+	}
+
```
rmontagnin: I think we can improve this code by making it a private function as follows: 

```go
// shouldUpdateDelegatorRewardsAmounts tells whether or not the delegatos rewards amounts should be updated at the given height 
func (m *Module) shouldUpdateDelegatorRewardsAmounts(height int64) bool {
	// Get interval cfg which decides when to execute "go m.refreshDelegatorsRewardsAmounts(b.Block.Height)"
	interval := m.cfg.RewardsFrequency
	if interval == 0 {
		log.Debug().Str("module", "distribution").Msg("delegator rewards refresh interval set to 0. Skipping refresh")
		return false
	}

	hasRewards, err := m.db.HasDelegatorRewards()
	if err != nil {
		log.Error().Str("module", "distribution").Err(err).Int64("height", b.Block.Height).
			Msg("error while checking delegators reward")
		return false
	}

	return !hasRewards || b.Block.Height%interval == 0
}
```

Then we can use this inside the `HandleBlock` method as such: 
```go
if m.shouldUpdateDelegatorRewardsAmounts(b.Block.Height) {
  go m.refreshDelegatosRewardsAmounts(b.Block.Height)
}
```

This would make the whole code more readable
```

Comment (huichiaotsou) [reply modules/distribution/handle_block.go]: got it

File: modules/distribution/utils_delegator_rewards.go [hunk 3]
```diff
@@ -54,6 +39,22 @@ func (m *Module) refreshDelegatorsRewardsAmounts(height int64) {
 	}
 }
 
+// shouldUpdateDelegatorRewardsAmounts tells whether or not the delegatos rewards amounts should be updated at the given height
+func (m *Module) shouldUpdateDelegatorRewardsAmounts(height int64) bool {
+	interval := m.cfg.RewardsFrequency
+	if interval == 0 {
+		log.Debug().Str("module", "distribution").Msg("delegator rewards refresh interval set to 0. Skipping refresh")
```
MonikaCat, line 46: Missing return value

modules/distribution/utils_delegator_rewards.go
```diff
@@ -54,6 +39,22 @@ func (m *Module) refreshDelegatorsRewardsAmounts(height int64) {
...
+	}
+
+	hasRewards, err := m.db.HasDelegatorRewards()
+	if err != nil {
+		log.Error().Str("module", "distribution").Err(err).Int64("height", height).
+			Msg("error while checking delegators reward")
```
MonikaCat, line 53: Missing return value here as well

modules/distribution/utils_delegator_rewards.go
huichiaotsou: fix

modules/distribution/utils_delegator_rewards.go
huichiaotsou: fix

MonikaCat:
LGTM

PR #225, opened 2021-10-08
fix: `MsgCreateValidator` and `MsgUpdateValidator` race condition

### LINKED ISSUE #219: `MSgCreateValidator` and `MsgUpdateValidator` race condition
## Bug description

Currently there is a possible bug while handling `MsgCreateValidator` and `MsgUpdateValidator` transactions that is due to a race condition. Suppose that a validator performs the following actions: 

1. At height `X`, it sends a `MsgCreateValidator` transaction with commission of 10% 
2. At height `X+10`, it sends a `MsgUpdateValidator` transaction updating other details but **not** the commission. 

Suppose also that the `MsgUpdateValidator` is parsed **before** the `MsgCreateValidator` transaction. What would happen is the following: 

1. BDJuno parses the `MsgUpdateValidator` and, since the commission is not set, it stores a commission of 0% 
2. BDJuno parses the `MsgCreateValidator` with the correct commission, but since the latest commission as a higher height, it is never updated and the correct commission (which should be 10%) is never stored. 

## Expected behavior

The `MsgCreateValidator` should store the correct infos, and the `MsgUpdateValidator` should not update the wrong fields.

## Possible solution
We might investigate whether is worth using some custom values to identify these cases. For example setting a commission of -1 for those validators who have not their commission set yet. Then the `MsgCreateValidator` will update those values accordingly.

rmontagnin:
## Description
This PR fixes the `MsgCreateValidator` and `MsgUpdateValidator` race condition bug by changing how we handle those types of messages. Instead of storing the infos directly from them, we use gRPC to query the entire data for the height that those msgs were transmitted to, and then store such information. This way even when a `MsgUpdateValidator` is sent, the entire data will be fetched and stored avoiding partial data storage that caused the bug. 

Fix #219 

## Checklist
- [ ] Wrote unit tests.  

### DIFF
--- a/database/staking_delegations_test.go
+++ b/database/staking_delegations_test.go
@@ -337,7 +337,7 @@ func (suite *DbTestSuite) TestDeleteCompletedRedelegations() {
 			srcValidator1.GetOperator(),
 			dstValidator1.GetOperator(),
 			sdk.NewCoin("cosmos", sdk.NewInt(100)),
-			time.Date(2021, 1, 1, 12, 00, 01, 000, time.UTC),
+			time.Date(2021, 1, 1, 12, 00, 00, 000, time.UTC),
 			10,
 		),
 		types.NewRedelegation(
@@ -537,7 +537,7 @@ func (suite *DbTestSuite) TestDeleteCompletedUnbondingDelegations() {
 			delegator1.String(),
 			validator1.GetOperator(),
 			sdk.NewCoin("cosmos", sdk.NewInt(100)),
-			time.Date(2021, 1, 1, 12, 00, 01, 000, time.UTC),
+			time.Date(2021, 1, 1, 12, 00, 00, 000, time.UTC),
 			10,
 		),
 		types.NewUnbondingDelegation(
--- a/modules/registrar.go
+++ b/modules/registrar.go
@@ -2,9 +2,10 @@ package modules
 
 import (
 	"fmt"
+	"os"
+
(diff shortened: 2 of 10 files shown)


database/staking_delegations_test.go
```diff
@@ -337,7 +337,7 @@ func (suite *DbTestSuite) TestDeleteCompletedRedelegations() {
 			srcValidator1.GetOperator(),
 			dstValidator1.GetOperator(),
 			sdk.NewCoin("cosmos", sdk.NewInt(100)),
-			time.Date(2021, 1, 1, 12, 00, 01, 000, time.UTC),
+			time.Date(2021, 1, 1, 12, 00, 00, 000, time.UTC),
```
rmontagnin, line 340: This fixes a failing test

database/staking_delegations_test.go
```diff
@@ -537,7 +537,7 @@ func (suite *DbTestSuite) TestDeleteCompletedUnbondingDelegations() {
 			delegator1.String(),
 			validator1.GetOperator(),
 			sdk.NewCoin("cosmos", sdk.NewInt(100)),
-			time.Date(2021, 1, 1, 12, 00, 01, 000, time.UTC),
+			time.Date(2021, 1, 1, 12, 00, 00, 000, time.UTC),
```
rmontagnin, line 540: Same as above

MonikaCat (approved):
utACK, LGTM

PR #226, opened 2021-10-08
fix: slashing events not being handled properly

### LINKED ISSUE #223: Validator slashing not updating delegations
## Bug description
Currently when a validator is slashed due to jail or other things, the associated delegations are not updated. We should make so  that they are, and also provide a command to fix all the currently wrong delegations too. 

## Steps to reproduce
 
1. Create a validator
2. Delegate some tokens to that validator
3. Make the validator slash due to downtime 
4. Observe the delegations

## Expected behavior

The delegations should update correctly

## Possible solution
Each block, observe to know which validators are now jailed/tombstoned. Then, update their delegations. 

Maybe there is even a way to read begin block events and know who has been slashed.

rmontagnin:
## Description
This PR fixes #223 by updating Juno to the latest version. With it, `HandleBlock` now has a new parameter that represents the result of the given block, which contains all the events emitted. From withing the `BeginBlockEvents` we can then read the `slash` events and get the addresses of those validator that are slashed. Once we have them, we simply update all the delegations for that height and validators. 

Tested with the following data:
- Chain: Desmos mainnet
- Parsing start height: `508780` (slashing should happen at block `508799`)

## Checklist
- [ ] Wrote unit tests.  

### DIFF
--- a/database/staking_delegations.go
+++ b/database/staking_delegations.go
@@ -109,6 +109,21 @@ func (db *Db) GetUserDelegationsAmount(address string) (sdk.Coins, error) {
 	return amount, nil
 }
 
+// DeleteValidatorDelegations removes all the delegations associated with the given validator consensus address
+func (db *Db) DeleteValidatorDelegations(valOperAddr string) error {
+	stmt := `
+DELETE FROM delegation USING validator_info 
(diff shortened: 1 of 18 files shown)


MonikaCat (approved):
LGTM

PR #228, opened 2021-10-10
feat: added fix command

### LINKED ISSUE #208: [LikeCoin] Proposal votes mismatch
## Bug description

Last checked likecoin's latest proposal had 69 votes but bdjuno was only able to log 9 of those. Would we be able to run a script that runs through the blocks and log only msgVote so that the proposal is correctly caught up?

### LINKED ISSUE #227: Add `fix` command
## Feature description 
Currently, when we fix something inside BDJuno it happens that the different users need to re-sync the whole database to make sure the data are parsed correctly. Instead, I think we can provide them with a `fix` command with various subcommands that allow to fix the issues we have solved. 

Some useful ones might be: 
- `fix staking validators` to fix the different validators delegations and information that might not have been parsed correctly due to the race condition (#219)
- `fix staking slashed-validators` to fix the delegations of the validators who have been slashed (#223)
- `fix ibc transactions` to fix all the IBC transactions that have not been parsed correctly inside v0.44.x (#218)

rmontagnin:
## Description
This PR implements the `fix` command that allows to fix BDJuno issues without having to re-index the whole database at once.

Closes #227 
Adds a way to fix #208 

## Checklist
- [ ] Targeted PR against correct branch.
- [ ] Wrote unit tests.  


MonikaCat (approved):
LGTM, utACK

huichiaotsou (approved):

PR #158, opened 2021-07-29
Vesting account data

rmontagnin:
## Description
Replaces #157. Originally opened by @HarleyAppleChoi. The base branch has changed named so we re-opened it 
close #138

## Checklist
- [x] Targeted PR against correct branch.
- [x] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [x] Wrote unit tests.  
- [x] Re-reviewed `Files changed` in the Github PR explorer.

PR #232, opened 2021-10-15
feat: add vesting account handling

### LINKED ISSUE #138: Vesting account data
## Feature description 

Big dipper + Forbole X might need to be able to display an account's vesting info if there are any. Is there an easy/ efficient way that we can log this data in bdjuno?

I was able to get the following data from

```
desmos q account desmos1snmetx9sfrmznr4z6d3lnwlyvwu36sgcpzzhlr
```

```
{
  "@type": "/desmos.profiles.v1beta1.Profile",
  "account": {
    "@type": "/cosmos.vesting.v1beta1.PeriodicVestingAccount",
    "base_vesting_account": {
      "base_account": {
        "address": "desmos1snmetx9sfrmznr4z6d3lnwlyvwu36sgcpzzhlr",
        "pub_key": {
          "@type": "/cosmos.crypto.secp256k1.PubKey",
          "key": "AsmSLYJM5CoIwuYQF+XvOFMSK1HeijFHF9XehSZGfET9"
        },
        "account_number": "143",
        "sequence": "3"
      },
      "original_vesting": [
        {
          "denom": "udaric",
          "amount": "30000000000"
        }
      ],
      "delegated_free": [],
      "delegated_vesting": [],
      "end_time": "1619766000"
    },
    "start_time": "1619506800",
    "vesting_periods": [
      {
        "length": "86400",
        "amount": [
          {
            "denom": "udaric",
            "amount": "10000000000"
          }
        ]
      },
      {
        "length": "86400",
        "amount": [
          {
            "denom": "udaric",
            "amount": "10000000000"
          }
        ]
      },
      {
        "length": "86400",
        "amount": [
          {
            "denom": "udaric",
            "amount": "10000000000"
          }
        ]
      }
    ]
  },
  "dtag": "Apollo",
  "moniker": "Apollo",
  "bio": "One of the most important and complex of the Greek gods, son of Zeus and Leto, and the twin brother of Artemis, goddess of the hunt.",
  "pictures": {
    "profile": "https://vignette.wikia.nocookie.net/olympians/images/5/54/Apollo_Viria.jpg/revision/latest/scale-to-width-down/298\\?cb\\=20180413123657",
    "cover": "https://upload.wikimedia.org/wikipedia/commons/thumb/e/e6/Hazy_Crazy_Sunrise.jpg/1200px-Hazy_Crazy_Sunrise.jpg"
  },
  "creation_date": "2021-04-30T09:31:51.619099537Z"
}
```

huichiaotsou:
## Description
This PR replaces #158 
close #138

## Checklist
- [x] Wrote unit tests.  


modules/auth/auth_accounts.go
```diff
@@ -32,6 +33,42 @@ func GetGenesisAccounts(appState map[string]json.RawMessage, cdc codec.Marshaler
...
+	if err := json.Unmarshal(appState[authttypes.ModuleName], &authState); err != nil {
+		return nil, err
+	}
+
+	// Store the accounts
+	accounts := []types.VestingAccount{}
+	for _, account := range authState.Accounts {
+		accountType := strings.Split(account.AccountType, ".")[3]
+
+		if accountType == "PeriodicVestingAccount" {
```
MonikaCat: We should store all types of vesting accounts, can you add them? 
-  ```BaseVestingAccount```
- ```ContinuousVestingAccount```
- ```DelayedVestingAccount```
- ```PeriodicVestingAccount```

database/auth.go
```diff
@@ -52,6 +53,41 @@ func (db *Db) saveAccounts(paramsNumber int, accounts []types.Account) error {
...
+			return err
+		}
+
+		params = append(params, account.Address, string(originalVestingBz), account.EndTime, account.StartTime, string(VestingPeriodsBz))
+	}
+
+	stmt = stmt[:len(stmt)-1]
+	stmt += `ON CONFLICT (address) DO NOTHING`
+
+	_, err := db.Sql.Exec(stmt, params...)
```
MonikaCat, line 133: I tested it on my local and i get an error:
```12:27PM ERR error while handling genesis err="error while storing genesis vesting accounts: pq: date/time field value out of range: \"1756649664\"" module=auth```

Comment (huichiaotsou) [reply database/auth.go:133]: oh yes I forgot to change the schema, it should be saved at TEXT. just pushed the fix now.

File: modules/auth/auth_accounts.go (hunk 1 above)
Comment (huichiaotsou) [reply modules/auth/auth_accounts.go]: sure, I will add them :D

File: database/schema/01-auth.sql [hunk 3]
```diff
@@ -1,4 +1,14 @@
...
 (
     address TEXT NOT NULL PRIMARY KEY
+);
+
+
+CREATE TABLE vesting_account
+(
+    address             TEXT        NOT NULL REFERENCES account (address) PRIMARY KEY,
+    original_vesting    JSONB       NOT NULL DEFAULT '{}',
+    end_time            TIMESTAMP   NOT NULL,
```
MonikaCat: ```start_time ``` and```end_time``` should be of type TEXT here, they are both strings: 

Screenshot 2021-10-15 at 12 52 19 PM

This will fix the _date/time field value out of range_ error ^

huichiaotsou: haha yes I just fixed them ^^

database/schema/01-auth.sql
```diff
@@ -1,4 +1,14 @@
 CREATE TABLE account
 (
     address TEXT NOT NULL PRIMARY KEY
+);
+
+
+CREATE TABLE vesting_account
+(
+    address             TEXT        NOT NULL REFERENCES account (address) PRIMARY KEY,
+    original_vesting    JSONB       NOT NULL DEFAULT '{}',
```
rmontagnin: I think it's better to use another type here. We should be able to use the `[]COINS` type instead of the `JSONB` to better support `sdk.Coins`.

database/schema/01-auth.sql
```diff
@@ -1,4 +1,14 @@
...
+    end_time            TEXT        NOT NULL,
+    start_time          TEXT        NOT NULL,
```
rmontagnin: Probably it's better to use a `TIMESTAMP WITHOUT TIMEZONE` instead of a `TEXT` here. That would allow for a better support inside GraphQL too

database/schema/01-auth.sql
```diff
@@ -1,4 +1,14 @@
...
+    vesting_periods     JSONB       NOT NULL DEFAULT '{}'
```
rmontagnin: I'm not a big fan of having a `JSONB` to represent a list of vesting periods since it's not possible to execute proper queries on this kind of type. Instead, I think we should create another table named `vesting_period` and insert those periods there. This would allow to query all the periods ending on a specific date or other things which might be useful for X or BigDipper

modules/auth/types.go
```diff
@@ -0,0 +1,31 @@
+package auth
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+// --- Types are defined according to genesis file structure ---
+
+type Accounts struct {
```
rmontagnin: Why are all these types defined here? What is their utility? Can't we simply use the Cosmos ones?

types/auth.go
```diff
@@ -11,3 +13,50 @@ func NewAccount(address string) Account {
 		Address: address,
 	}
 }
+
+// --------------- For Vesting Account ---------------
+
+// Account represents a chain account
+type VestingAccount struct {
```
rmontagnin: As above, why are these re-declared since they are already available inside the SDK?

huichiaotsou: Hi @RiccardoM , I added the types on the side because I couldn't really get the account type with Codec.. so I did with brute force to parse them into json. Do you know a way how I can parse/unpack the accounts?

database/schema/01-auth.sql
huichiaotsou: got it

database/schema/01-auth.sql
huichiaotsou: Sure will modify it

types/auth.go
rmontagnin: @huichiaotsou In order to understand how to unpack accounts, it's worth to understand how they are handled from Cosmos and why they are packed in a strange way. The best way to do this is though a series of considerations: 

1. All the data returned from a gRPC call is serialized using [Protobuf](https://developers.google.com/protocol-buffers)
2. In order to represent a generic account, inside the Cosmos SDK there are multiple Go interfaces used. The most common are: 
   1. `AccountI`, to represent a generic account
   2. `VestingAccount` to represent a generic vesting account 
3. Protobuf does **not** support Go interfaces by default. For this reason, the Cosmos team has decided to use the [`Any`](https://developers.google.com/protocol-buffers/docs/proto3#any) type to serialize them. 

So, in order to read the details of a vesting account you need to do the following: 

1. Unpack that account from `Any` into an `AccountI` interface: 
```go
var accountI authtypes.AccountI
err := cdc.UnpackAny(account, &accountI)
if err != nil {
  return err
}
```

2. Check if that `accountI` is also implementing the `VestingAccountI` interface: 
```go
vestingAccount, ok := accountI.(exported.VestingAccount)
if !ok {
  // The account is not a VestingAccount. Do as you wish (skip the for loop, return an error, ...)
}
```

From there, you can use `vestingAccount` however you wish since it will be of type `VestingAccount`. 

#### Side note 
Since they can be of various types ([`ContinousVestingAccount`](https://github.com/cosmos/cosmos-sdk/blob/master/proto/cosmos/vesting/v1beta1/vesting.proto#L34), [`DelayedVestingAccount`](https://github.com/cosmos/cosmos-sdk/blob/master/proto/cosmos/vesting/v1beta1/vesting.proto#L45) and [`PeriodicVestingAccount`](https://github.com/cosmos/cosmos-sdk/blob/master/proto/cosmos/vesting/v1beta1/vesting.proto#L63)), the best way to store `VestingAccount` instances is to simply serialize them as JSON and store them inside a table. However, this might make it really hard to get some useful insights that we might want to display on BigDipper (such as total vested/unvested tokens, next unvesting period, etc). 

For this reason, I would like to ask @ryuash and @calvinkei how they are going to use the vesting data in order to decide how to store them accordingly.

ryuash: By default Big Dipper 2.0 will not be using this data in favor of Forbole X. 
Inside the Forbole X wallet details page I would assume it would show xx tokens are locked until xx (maybe a countdown type of animation)

Maybe @calvinkei can clarify

rmontagnin: @ryuash Wouldn't it be interesting for BigDipper to show a global amount of locked/unlocked tokens? Asking @kwunyeung also. 

That might be interesting for users, no? To know when the next large unlock will happen or other interesting things

calvinkei: This is the UI for showing vesting tokens on forbole x
Screenshot 2021-10-18 at 3 30 30 PM

So, it needs to support "total vesting tokens" and pagination. Probably filter & sort by date as well

ryuash: Hahs, if we do some feature like that then we definitely cannot throw a JSONB in to psql.

I would need to get all possible vesting periods in the genesis file and query by xx time.

rmontagnin: @huichiaotsou  In order to preserve forward-compatibility with new features that might come up I would suggest then adding a better handling for different vesting account types. We can do so by doing this: 

```golang
// inside handler
db.StoreVestingAccount(vestingAccount)

// -------------------------

func (db *Database) StoreVestingAccount(vestingAccount exported.VestingAccount) error {
  if continousVestingAccount, ok := vestingAccount.(*vestingtypes.ContinousVestingAccount); ok {
    return db.storeContinousVestingAccount(continousVestingAccount)
  }

  if delayedVestingAccount, ok := vestingAccount.(*vestingtypes.DelayedVestingAccount); ok {
    return db.storeDelayedVestingAccount(continousVestingAccount)
  }

  if periodicVestingAccount, ok := vestingAccount.(*vestingtypes.PeriodicVestingAccount); ok {
    return db.storePeriodicVestingAccount(continousVestingAccount)
  }

  return fmt.Errorf("invalid vesting account type: %T", vestingAccount")
```

This way we can handling the storing of different types properly, with various periods etc

kwunyeung: @RiccardoM I'm also interested in showing vesting amount on BD but I do agree with @ryuash that it could be in favour of Forbole X at some point if BD not showing them by default. I think we can have this feature showing on X first. I also wanna see if there will be any feedback or request on getting this feature on BD.

Agree not storing the vesting periods in `JSONB` as the structure is very clear that each record has an array of coins and a vested date. It's not any arbitrary json content where we may search in certain keys inside the json.

MonikaCat (approved):
Tested, LGTM

database/auth.go
```diff
@@ -2,13 +2,18 @@ package database
...
+	vestingtypes "github.com/cosmos/cosmos-sdk/x/auth/vesting/types"
+	dbtypes "github.com/forbole/bdjuno/v2/database/types"
 	dbutils "github.com/forbole/bdjuno/v2/database/utils"
+	"github.com/lib/pq"
 
 	"github.com/forbole/bdjuno/v2/types"
 )
 
-// SaveAccounts saves the given accounts inside the database
+// SaveAccounts saves the given account addresses inside the database
```
rmontagnin: I think it's fine to leave the description as it was before. We might never know in the future if we want to add other things to each account honestly. 
```suggestion
// SaveAccounts saves the given accounts inside the database
```

database/auth.go
```diff
@@ -52,6 +57,146 @@ func (db *Db) saveAccounts(paramsNumber int, accounts []types.Account) error {
 	return nil
 }
 
+// SaveVestingAccounts saves the given vesting account details inside the database, including:
+// type, address, original vesting, endTime, startTime, and vesting periods
```
rmontagnin: No need to specify what is saved here: 
```suggestion
// SaveVestingAccounts saves the given vesting accounts inside the database
```

database/auth.go
```diff
@@ -52,6 +57,146 @@ func (db *Db) saveAccounts(paramsNumber int, accounts []types.Account) error {
...
+	}
+
+	for _, account := range vestingAccounts {
+		switch vestingAccount := account.(type) {
+		case *vestingtypes.ContinuousVestingAccount:
+			ContinuousVestingAccount := types.NewContinuousVestingAccount(*vestingAccount)
+			err := db.storeContinuousVestingAccount(ContinuousVestingAccount)
+			if err != nil {
+				return fmt.Errorf("error while storing Continuous Vesting Account: %s", err)
+			}
```
rmontagnin: There are a couple of improvements that can be applied to all the following `cases`. I will use the first one as a reference, but the same apply to the other two as well. 

First, there is no need to do this: 
```go
ContinuousVestingAccount := types.NewContinuousVestingAccount(*vestingAccount)
err := db.storeContinuousVestingAccount(ContinuousVestingAccount)
```
Since Go is able to implicitly determine the type of a variable based on the `case` you are using, this means that after line 70 the `account` variable will now have type `vestingtypes.ContinuousVestingAccount`. Creating a new `types.ContinuousVestingAccount` instance on top of it it's only redundant and might make it harder to handler later. 

Also, Go follows the `camelCase` convention for variables, not the `PascalCase` one. So here it should be `continousVestingAccount` and not `ContinuousVestingAccount`. 

Anyway, it would be a lot simpler here to do the following instead:
```go
case *vestingtypes.ContinuousVestingAccount:
	err := db.storeContinuousVestingAccount(account)
```

And then define the `storeContinuousVestingAccount` method as follows: 
```go
func (db *Db) storeContinuousVestingAccount(account *vestingtypes.ContinuousVestingAccount) error {
```

database/auth.go
```diff
@@ -52,6 +57,146 @@ func (db *Db) saveAccounts(paramsNumber int, accounts []types.Account) error {
...
+		account.Address,
+		pq.Array(dbtypes.NewDbCoins(account.OriginalVesting)),
+		time.Unix(account.EndTime, 0).Format(time.RFC3339),
+		time.Unix(account.StartTime, 0).Format(time.RFC3339),
+	)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
rmontagnin: From what I can tell, this method is doing the same exact things as `storeDelayedVestingAccount` and `storePeriodicVestingAccount`. For this reason, I think it can be simplified as follows: 

```golang
// storeContinuousVestingAccount stores the vesting account details of type ContinuousVestingAccount into the database
func (db *Db) storeVestingAccount(account exported.VestingAccount) error {
	stmt := `
INSERT INTO vesting_account (type, address, original_vesting, end_time, start_time) 
VALUES ($1, $2, $3, $4, $5)
ON CONFLICT (address) DO UPDATE 
	SET original_vesting = excluded.original_vesting, 
	    end_time = excluded.end_time, 
	    start_time = excluded.start_time`

	_, err := db.Sql.Exec(stmt,
		proto.MessageName(account),
		account.GetAddress(),
		pq.Array(dbtypes.NewDbCoins(account.GetOriginalVesting())),
		time.Unix(0, account.GetEndTime()),
		time.Unix(0, account.GetStartTime()),
	)
	if err != nil {
		return err
	}
	return nil
}
```

As you might see, there are a couple of changes here: 
**1.** We are using `GetAddress()`, `GetOriginalVesting()`, `GetEndTime()` and `GetStarTime()` instead of the account fields directly.  
This is due to the fact that the [`VestingAccount`](https://github.com/cosmos/cosmos-sdk/blob/master/x/auth/vesting/exported/exported.go#L12) interface implements those methods and so we can use them safely without worrying about how they are implemented. 

**2.** We are using `proto.MessageName(account)` as the account type.  
This is possible thanks to the fact that `VestingAccount` extends `AccountI` which extends `proto.Message` and so we can safely get the type using that method. The full docs is [here](https://pkg.go.dev/github.com/gogo/protobuf/proto#MessageName).

**3.**. We should be using `time.Unix(sec, nsec)` appropriately. Since the `GetEndTime` and `GetStartTime` both return a nanosecond value, we should use that as the second parameter of `Unix` rather than the first one.

**4.** There's no need to call `Format(time.RFC3339)` since the PostgreSQL Go driver can take care of `time.Time` values properly anyway.

database/schema/01-auth.sql
```diff
@@ -1,4 +1,34 @@
...
+/* ---- start_time can be empty on DelayedVestingAccount ---- */
+
+CREATE UNIQUE INDEX vesting_account_address_idx ON vesting_account (address);
+
+
+CREATE TABLE vesting_period
+(
+    vesting_account_id  INT     NOT NULL REFERENCES vesting_account (id),
+    period_order        INT     NOT NULL,
+    length              TEXT    NOT NULL,
```
rmontagnin: I think it's much better to store the `length` and a `BIGINT` rather than a `TEXT` here

types/auth.go
```diff
@@ -11,3 +16,65 @@ func NewAccount(address string) Account {
 		Address: address,
 	}
 }
+
+// --------------- For Vesting Accounts ---------------
+
+// ContinuousVestingAccount represents a Continuous Vesting Account
+type ContinuousVestingAccount struct {
```
rmontagnin: This and all the below types should be deleted

database/auth.go
huichiaotsou: fixed

database/auth.go
huichiaotsou: fixed

database/auth.go
huichiaotsou: fixed

types/auth.go
huichiaotsou: fixed

database/auth.go
huichiaotsou: fixed

database/schema/01-auth.sql
huichiaotsou: fixed

database/auth.go
huichiaotsou: I think the **account.GetEndTime()** and **account.GetStartTime()** return seconds so I will switch their positions in time.Unix() :D

also I added **.String()** to account.GetAddress()

database/types/auth.go
```diff
@@ -16,3 +20,144 @@ func NewAccountRow(address string) AccountRow {
 func (a AccountRow) Equal(b AccountRow) bool {
 	return a.Address == b.Address
 }
+
+// --------------- For Vesting Accounts ---------------
+
+// ContinuousVestingAccountRow represents a single row inside the vesting_account table
+type ContinuousVestingAccountRow struct {
```
rmontagnin: Do we still need all of these custom rows types?

database/auth.go
```diff
@@ -52,6 +58,92 @@ func (db *Db) saveAccounts(paramsNumber int, accounts []types.Account) error {
...
+		time.Unix(account.GetStartTime(), 0),
+	).Scan(&vestingAccountRowID)
+
+	if err != nil {
+		return fmt.Errorf("error while saving Vesting Account: %s", err)
+	}
+
+	if periodicVestingAccount, ok := account.(*vestingtypes.PeriodicVestingAccount); ok {
+		return db.storeVestingPeriods(vestingAccountRowID, periodicVestingAccount.VestingPeriods)
+	}
```
rmontagnin: I think this can be moved inside the `SaveVestingAccounts` method instead in order to avoid having to re-perform the casting

database/auth.go
```diff
@@ -52,6 +58,92 @@ func (db *Db) saveAccounts(paramsNumber int, accounts []types.Account) error {
...
+
+	for _, account := range vestingAccounts {
+		switch account.(type) {
+		case *vestingtypes.ContinuousVestingAccount:
+			err := db.storeVestingAccount(account)
+			if err != nil {
+				return fmt.Errorf("error while storing Continuous Vesting Account: %s", err)
+			}
+
+		case *vestingtypes.DelayedVestingAccount:
```
rmontagnin: These can be simplified to be a two-in-one case I believe (not sure though):
```go
	case *vestingtypes.ContinuousVestingAccount:
	case *vestingtypes.DelayedVestingAccount:
```

database/auth.go
rmontagnin: Ok great!

database/types/auth.go
huichiaotsou: i left them for writing unit tests later hehe, I think I will delete them for the moment since the code structures are different now

database/auth.go
huichiaotsou: fixed

database/auth.go
huichiaotsou: fixed

database/auth.go
```diff
@@ -52,6 +58,83 @@ func (db *Db) saveAccounts(paramsNumber int, accounts []types.Account) error {
...
+			if err != nil {
+				return err
+			}
+
+		case *vestingtypes.PeriodicVestingAccount:
+			vestingAccountRowID, err := db.storeVestingAccount(account)
+			if err != nil {
+				return err
+			}
+			db.storeVestingPeriods(vestingAccountRowID, vestingAccount.VestingPeriods)
```
rmontagnin: We need to return the error here

```suggestion
			return db.storeVestingPeriods(vestingAccountRowID, vestingAccount.VestingPeriods)
```

modules/auth/handle_genesis.go
```diff
@@ -11,17 +11,28 @@ import (
 
 // HandleGenesis implements modules.GenesisModule
 func (m *Module) HandleGenesis(_ *tmtypes.GenesisDoc, appState map[string]json.RawMessage) error {
-	log.Debug().Str("module", "auth").Msg("parsing genesis")
```
rmontagnin, line 14: Can we revert this? 
```suggestion
	log.Debug().Str("module", "auth").Msg("parsing genesis")
```

modules/auth/handle_genesis.go
```diff
@@ -11,17 +11,28 @@ import (
...
 
+	// Handle account addresses
+	log.Debug().Str("module", "auth").Msg("parsing genesis")
```
rmontagnin: And these as well?
```suggestion
```

modules/auth/handle_genesis.go
```diff
@@ -11,17 +11,28 @@ import (
...
 		return fmt.Errorf("error while getting genesis accounts: %s", err)
 	}
-
 	err = m.db.SaveAccounts(accounts)
 	if err != nil {
 		return fmt.Errorf("error while storing genesis accounts: %s", err)
 	}
 
+	// Handle storing vesting accounts
+	log.Debug().Str("module", "auth/vesting").Msg("parsing genesis")
```
rmontagnin: I think we can remove this
```suggestion
```

rmontagnin (approved):
utACK. Can I have a tACK from @MonikaCat please?

database/auth.go
```diff
@@ -52,6 +58,83 @@ func (db *Db) saveAccounts(paramsNumber int, accounts []types.Account) error {
...
+			if err != nil {
+				return err
+			}
+
+		case *vestingtypes.PeriodicVestingAccount:
+			vestingAccountRowID, err := db.storeVestingAccount(account)
+			if err != nil {
+				return err
+			}
+			return db.storeVestingPeriods(vestingAccountRowID, vestingAccount.VestingPeriods)
```
MonikaCat: It's better if we remove ```return``` and call the function here directly, otherwise bdjuno will only process the first account on the list.

rmontagnin: Oh that's true, my bad. Then we should definitely add an error checking to make sure the returned error is not null

MonikaCat: Sure, I updated the code!

MonikaCat:
tACK, LGTM

Issue #216, opened 2021-10-04
Blocks and transactions are skipped sometimes


kwunyeung:
## Bug description

This happens on LikeCoin BD. The prop#16 was passed. There were 81 vote transactions for prop#16 but only 8 votes were indexed in the database.

https://likecoin.bigdipper.live/proposals/16

Blocks are missing sometimes. There are reports on BitCanna.

![image.png](https://images.zenhubusercontent.com/610d12dcbed147b45e407c24/9ab29e4c-dc53-491f-a710-aa808918b7c6)

Possible to avoid missing blocks being indexed? How about if a block is failed to be indexed at some point, `BDJuno` would retry?

kwunyeung:
Transaction issue is mentioned in #208 as well.

rmontagnin:
> Possible to avoid missing blocks being indexed? How about if a block is failed to be indexed at some point, `BDJuno` would retry?

I don't understand what you mean with this. If BDJuno doesn't catch some blocks, they won't be indexed. 

And if there are some errors while parsing the blocks, it will log those errors and try again later until the block is parsed.

MonikaCat:
Closed with #248

PR #248, opened 2021-10-25
fix: add refetch blocks and transactions fix command

MonikaCat:
## Description

References #216 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


cmd/fix/blocks/blocks.go
```diff
@@ -0,0 +1,120 @@
...
+		return fmt.Errorf("error while updating block %v: %s", blockHeight, err)
+	}
+
+	return nil
+}
+
+func refreshTxs(parseCtx *parse.Context, consensusModule *consensus.Module, block *tmctypes.ResultBlock) error {
+
+	for _, tx := range block.Block.Txs {
+		var txDetails *juno.Tx
```
huichiaotsou: I think we can remove the variable declaration at line 106 because declaration is made at line 108

modules/consensus/utils.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// GetStartingHeight takes starting height value from the config file
+func (m *Module) GetStartingHeight() int64 {
+	startHeight := m.cfg.StartHeight
+	return startHeight
+}
+
+// UpdateTxs updates txs in database
+func (m *Module) HandleMessages(txDetails *types.Tx) error {
```
huichiaotsou: The comment needs updating :p

database/consensus.go
```diff
@@ -184,3 +193,145 @@ func (db *Db) GetGenesis() (*types.Genesis, error) {
...
+
+// CheckIfBlockIsMissing checks if block is already stored in database
+func (db *Db) CheckIfBlockIsMissing(height int64) bool {
+	var block []dbtypes.BlockRow
+	stmt := `SELECT * FROM block WHERE height = $1`
+
+	err := db.Sqlx.Select(&block, stmt, height)
+
+	return err != nil
+}
```
huichiaotsou: we should check "len(block) == 0" instead of checking "error != nil", otherwise it would always return false (the err would still be nil even when the selected row is empty)

MonikaCat: Ok I added the checks back to the code 👍🏻

huichiaotsou (approved):
LGTM, tested locally

database/consensus.go
```diff
@@ -184,3 +193,151 @@ func (db *Db) GetGenesis() (*types.Genesis, error) {
...
+	}
+	if len(block) != 0 {
+		return false
+	}
+
+	return true
+}
+
+// UpdateBlocksInDatabase updates given block in database
+func (db *Db) UpdateBlockInDatabase(block *tmctypes.ResultBlock, blockResults *tmctypes.ResultBlockResults) error {
```
rmontagnin: Why do we need to re-create all these methods that are already part of the `Database` interface? Can't we simply use `db.SaveBlock`, `db.SaveTransaction` and `db.SaveMessage` instead?

database/staking_delegations.go
```diff
@@ -27,6 +29,20 @@ func (db *Db) SaveDelegations(delegations []types.Delegation) error {
...
+			if scanErr != nil {
+				return fmt.Errorf("error while scanning error: %s", scanErr)
+			}
+
+			// If error is different than delegation not found, we need to return it
+			if errorCode != codes.NotFound.String() {
+				return fmt.Errorf("error while searching for delegations %s", err)
+			}
+		}
+
```
rmontagnin: Why is this necessary?

modules/consensus/config.go
```diff
@@ -0,0 +1,29 @@
+package consensus
+
+import "gopkg.in/yaml.v3"
+
+// Config contains the configuration about chain start height
+type Config struct {
+	StartHeight int64 `yaml:"start_height,omitempty"`
+}
```
rmontagnin: It's not necessary to define a new structure for this thing honestly. We can simply get the config that is parsed when the `bdjuno` command is used and get the starting height from there: 

```go
// Get the config
startHeight := config.Cfg.ParserStartHeight
```

database/consensus.go
MonikaCat: Sure, I updated the code!

modules/consensus/config.go
MonikaCat: Updated

database/staking_delegations.go
MonikaCat: I added this to handle the error that bdjuno throws when it parses old delegate type tx and can't find it

rmontagnin: I don't think we should ignore this error honestly. If the validator is not stored, we can't ignore the fact that something is wrong. Something like this should never happen

MonikaCat: Okay sure! I've tested it locally and I can't reproduce this error anymore so I have removed this part from the code

cmd/fix/blocks/blocks.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/forbole/bdjuno/v2/utils"
+	"github.com/forbole/juno/v2/cmd/parse"
+	junomessages "github.com/forbole/juno/v2/modules/messages"
+	"github.com/forbole/juno/v2/types/config"
+	"github.com/spf13/cobra"
+	tmctypes "github.com/tendermint/tendermint/rpc/core/types"
+)
+
+// blocksCmd returns a Cobra command that allows to fix missing blocks in database
+func blocksCmd(parseConfig *parse.Config) *cobra.Command {
```
rmontagnin, line 14: I think there is an easier way to implement this by using the existing `Worker` structure: 
1. Get the parsing context as you do
2. Build a new `Worker` instance
3. For each height, call the `worker.Process(height)` method returning any error

This will make sure that we call everything needed even if we add new modules in the future and would allow us to remove all the code duplication added to store things inside the database. 

**Note**: to implement this the currently private `process` method must be made public inside Juno.

MonikaCat: Great! I updated the code and exported `process` method in juno repository

database/staking_delegations.go
```diff
@@ -56,7 +56,7 @@ INSERT INTO delegation (validator_address, delegator_address, amount, height) VA
 		// Get the validator consensus address
 		consAddr, err := db.GetValidatorConsensusAddress(delegation.ValidatorOperAddr)
 		if err != nil {
-			return fmt.Errorf("error while gettting validator consensus address: %s", err)
+			return err
```
rmontagnin: Can we revert this?

modules/staking/utils_delegations.go
```diff
@@ -49,7 +45,7 @@ func (m *Module) getValidatorDelegations(height int64, validator string) ([]type
 func (m *Module) getDelegatorDelegations(height int64, delegator string) ([]types.Delegation, error) {
 	responses, err := m.source.GetDelegatorDelegations(height, delegator)
 	if err != nil {
-		return nil, fmt.Errorf("error while getting delegator delegations: %s", err)
```
rmontagnin: Can we revert this?

modules/staking/utils_delegations.go
```diff
@@ -85,7 +81,7 @@ func (m *Module) refreshDelegatorDelegations(height int64, delegator string) err
 	if err != nil {
 		// Get the error code
 		var code string
-		_, scanErr := fmt.Sscanf(err.Error(), ErrDelegationNotFound, &code, &code)
```
rmontagnin: Can we revert this as well?

modules/utils/errors.go
```diff
@@ -4,6 +4,8 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+const ErrNotFound = "rpc error: code = %s desc = rpc error: code = %s desc ="
```
rmontagnin: No need to put this here since it's not used anywhere else than the delegations

rmontagnin (approved):

PR #252, opened 2021-10-27
feat: handle update params proposals in gov module

### LINKED ISSUE #194: Handle update params proposals
## Feature description 
We should handle the governance proposals that update parameters properly, refreshing such parameters once the proposals pass. 

## Implementation proposal
We can simply check for different proposals each time and, if a proposal has passed, simply update the parameters (if they have not been already).

huichiaotsou:
## Description

Closes: #194 

This PR centralizes the handling of ParameterChangeProposal in gov module and updates the params accordingly(for gov, staking, slashing, mint, and distribution).

The open ParameterChangeProposals are selected from DB and are checked if any of them passed, the params of corresponding module are updated. The UpdateParams methods originally called in Handle_block of above modules are removed.

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

- [x] confirmed the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [x] reviewed API design and naming
- [x] reviewed documentation is accurate
- [x] reviewed tests and test coverage


types/gov.go
```diff
@@ -149,6 +151,38 @@ func NewProposalUpdate(
 	}
 }
 
+// ParameterChangeProposal represents a single governance ParameterChangeProposal
+type ParameterChangeProposal struct {
```
rmontagnin: Why is this re-declared here? Can't use the `govtypes.ParamChangeProposal` instead?

modules/gov/utils_proposal.go
```diff
@@ -66,6 +69,53 @@ func (m *Module) UpdateProposal(height int64, blockVals *tmctypes.ResultValidato
...
+			return m.UpdateParams(height)
+		case slashingtypes.ModuleName:
+			return m.slashingModule.UpdateParams(height)
+		case stakingtypes.ModuleName:
+			return m.stakingModule.UpdateParams(height)
+		case minttypes.ModuleName:
+			return m.mintModule.UpdateParams(height)
+		case distrtypes.ModuleName:
+			return m.distrModule.UpdateParams(height)
+		}
```
rmontagnin: This will not work properly if there are multiple changes inside the same `ParamChangeProposal`. Instead, what we should do is: 

1. Iterate through all the changes and get a list of subspaces to be updated
2. For each subspace, call the method and return **only** if `err != nil`

Otherwise we will return even if `err == nil` thus never updating all the subspaces

modules/gov/handle_block.go
```diff
@@ -8,52 +8,42 @@ import (
...
 	}
-
 	return nil
 }
 
-// updateParams updates the governance parameters for the given height
-func (m *Module) updateParams(height int64) error {
-	depositParams, err := m.source.DepositParams(height)
+// updateParamChangeProposals updates the params if a ParamChangeProposal passed
+func (m *Module) handleParamChangeProposals(height int64) error {
```
rmontagnin: The comment should start with the method name

types/gov.go
huichiaotsou: I declared a customized ParameterChangeProposal with ProposalID because the original `proposaltypes.ParamChangeProposal` does not have it, but it is required for m.source.Proposal(height, id)

modules/gov/utils_proposal.go
huichiaotsou: 1. The `changes` are iterated at line 101, so should be ok? when each change matches the module name, the corresponding `UpdatParams` will be called. Or you mean there can be several `Subspace` in 1 change of the `changes` array ?

2. Will fix the error handling :D

modules/gov/utils_proposal.go
```diff
@@ -66,6 +69,68 @@ func (m *Module) UpdateProposal(height int64, blockVals *tmctypes.ResultValidato
...
+			}
+		case distrtypes.ModuleName:
+			err := m.distrModule.UpdateParams(height)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
```
huichiaotsou: Hi @RiccardoM , is it like this?

rmontagnin: @huichiaotsou Yes, looks great

modules/gov/utils_proposal.go
```diff
@@ -66,6 +69,68 @@ func (m *Module) UpdateProposal(height int64, blockVals *tmctypes.ResultValidato
 	return nil
 }
 
+func (m *Module) UpdateParamsFromParamChangeProposal(height int64, id uint64, proposal types.ParameterChangeProposal) error {
```
rmontagnin: I think we can move most of the code inside this function inside the `UpdateProposal` method actually, that might simplify things a lot. We simply have to check the content type and then call `updateModuleParams` in case

huichiaotsou: that's true! I have integrated them :D

modules/mint/handle_block.go
```diff
@@ -3,40 +3,12 @@ package mint
...
 	juno "github.com/forbole/juno/v2/types"
 
-	"github.com/rs/zerolog/log"
 	tmctypes "github.com/tendermint/tendermint/rpc/core/types"
-
-	"github.com/forbole/bdjuno/v2/types"
 )
 
 // HandleBlock implements modules.BlockModule
 func (m *Module) HandleBlock(
```
MonikaCat: Since we don't need block handling for mint module anymore, should we remove it?

modules/registrar.go
```diff
@@ -120,7 +122,7 @@ func (r *Registrar) BuildModules(ctx registrar.Context) jmodules.Modules {
 		bankModule,
 		consensus.NewModule(db),
 		distrModule,
-		gov.NewModule(cdc, sources.GovSource, authModule, bankModule, stakingModule, db),
+		gov.NewModule(cdc, sources.GovSource, authModule, bankModule, stakingModule, slashingModule, mintModule, distrModule, db),
```
MonikaCat: Can you update modules to be in alphabetical order?

huichiaotsou: ok :D

modules/mint/handle_block.go
huichiaotsou: removed ^^

MonikaCat (approved):
LGTM

rmontagnin (approved):

PR #257, opened 2021-11-08
feat: add CMD parse-genesis

### LINKED ISSUE #246: Add command to parse genesis
Issue:
Lots of third party bdjunos are never setup correctly, mostly an issue with parsing genesis.

Solution:
Remove parse genesis from`bdjuno init` and run it in a separate cmd.
Have the user run `bdjuno parse-genesis` first before `bdjuno init` allowing users to know early on if the genesis was parsed correctly or not before proceeding.

huichiaotsou:
## Description

Closes: #246 
Note: forbole/juno#30 and forbole/juno#31 will have to be merged 

Add new command **bdjuno parse-genesis** to parse the genesis.json file 
- users can specify which module in the genesis they would like to parse
- re-implemented HandleGenesis of staking mod to be run without a running node
- removed log.Debug() from HandleGenesis methods as **parse-genesis** is executed off a running node, avoiding log like: 
```
{"level":"debug","module":"auth","time":"2021-11-16T15:28:37+08:00","message":"parsing genesis"}
{"level":"debug","module":"bank","time":"2021-11-16T15:28:39+08:00","message":"parsing genesis"}
```
- invalid arguments of module name are printed after the genesis is parsed
- Update doc: forbole/big-dipper-2.0-documentation#16
---

### Author Checklist

I have...

### Reviewers Checklist

I have...


rmontagnin:
Before reviewing this, can you please fix the linting and tests @huichiaotsou? Also, when ready, can you mark this PR as ready for review?

huichiaotsou:
> Before reviewing this, can you please fix the linting and tests @huichiaotsou? Also, when ready, can you mark this PR as ready for review?

Hi @RiccardoM , I have changed it to ready for review. For the linter and test errors, it's because in [juno](https://github.com/forbole/juno) I did 2 slight modification in order to separate parse cmd and parse genesis cmd, so I think those PR should be reviewed and merged first.

MonikaCat (approved):
LGTM, tested locally with #266

cmd/parse-genesis/parseGenesis.go
```diff
@@ -0,0 +1,43 @@
...
+		Short:   "Parse genesis file. To parse specific modules, input module names as arguments",
+		Example: "bdjuno parse-genesis auth bank consensus gov history staking",
+		PreRunE: parse.ReadConfig(parseCfg),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			genesisDoc, genesisState, err := utils.GetGenesisDocAndState()
+			if err != nil {
+				return fmt.Errorf("error while getting genesis doc or state: %s", err)
+			}
+
+			registeredModules, err := utils.GetRegisteredModules(parseCfg)
```
rmontagnin: Instead of defining a new function that mimics the behavior of an already existing function, we should leverage the code that already exists instead. In this case, the `GetREgisteredModules` function can be easily replaced with the following lines: 
```go
parseCtx, err := parse.GetParsingContext(parseCfg)
if err != nil {
    return err
}
```

Looking at the value returned from `GetParsingContext`, we can see that the returned `Context` object already contains a list of enabled modules: 
```go
type Context struct {
    EncodingConfig *params.EncodingConfig
    Node           node.Node
    Database       database.Database
    Logger         logging.Logger
    Modules        []modules.Module
}
```

Once we have the context, we can then simply do: 
```go
for _, module := range modules {
    if genesisModule, ok := module.(modules.GenesisModule) { 
        genesisModule.ParseGenesis(...)
    }
}
```

cmd/parse-genesis/utils/utils.go
```diff
@@ -0,0 +1,105 @@
...
+	// Get the modules
+	context := modsregistrar.NewContext(cfg, sdkConfig, &encodingConfig, db, nil, parseConfig.GetLogger())
+	mods := parseConfig.GetRegistrar().BuildModules(context)
+	registeredModules := modsregistrar.GetModules(mods, cfg.Chain.Modules, parseConfig.GetLogger())
+
+	return registeredModules, nil
+}
+
+// GetGenesisDocAndState generates and returns the genesis doc and genesis state with genesis.json file
+func GetGenesisDocAndState() (*tmtypes.GenesisDoc, map[string]json.RawMessage, error) {
```
rmontagnin: Currently this function is completely ignoring the fact that the user might have selected to not use a genesis file but rather fetch the genesis directly from the node instead. For this reason, the overall behavior of `GetGenesisDocAndState` should be the following: 

1. Look at the config `GenesisFilePath` value specified. 
2. If the path is **empty**, fetch the genesis doc from the node. 
3. If the path is **not empty**, read the genesis from the file. 
4. In both cases, get the genesis state from the genesis doc

I also think that since this method might be useful elsewhere, it might be better to implement it inside the Juno library instead of here. Currently Juno uses the following code to get the genesis doc: 

```go
cfg := config.Cfg.Parser

var genesis *tmtypes.GenesisDoc
if strings.TrimSpace(cfg.GenesisFilePath) != "" {
    genesis, err = w.getGenesisFromFilePath(cfg.GenesisFilePath)
    if err != nil {
        return err
    }
} else {
    genesis, err = w.getGenesisFromRPC()
    if err != nil {
        return err
    }
}


// ---

// getGenesisFromRPC returns the genesis read from the RPCConfig endpoint
func (w Worker) getGenesisFromRPC() (*tmtypes.GenesisDoc, error) {
    w.logger.Debug("getting genesis from RPCConfig")

    response, err := w.node.Genesis()
    if err != nil {
        return nil, fmt.Errorf("failed to get genesis: %s", err)
    }

    return response.Genesis, nil
}

// getGenesisFromFilePath tries reading the genesis doc from the given path
func (w Worker) getGenesisFromFilePath(path string) (*tmtypes.GenesisDoc, error) {
    w.logger.Debug("reading genesis from file", "path", path)

    bz, err := tmos.ReadFile(path)
    if err != nil {
        return nil, fmt.Errorf("failed to read genesis file: %s", err)
    }

    var genDoc tmtypes.GenesisDoc
    err = tmjson.Unmarshal(bz, &genDoc)
    if err != nil {
        return nil, fmt.Errorf("failed to unmarshal genesis doc: %s", err)
    }

    return &genDoc, nil
}
```

Instead I think we can make it more general, something like this: 
```go
func GetGenesisDocAndState(parserCfg parserconfig.Config, node node.Node) (*tmtypes.GenesisDoc, map[string[json.RawMessage], error) {
    ...
}
```

This way this function can then be used also by BDJuno as well as other libraries using Juno.

rmontagnin (commented):
I think there is a lot of code duplication here that is not necessary

cmd/parse-genesis/parseGenesis.go
huichiaotsou: got it. will use parseCtx instead to get the registered modules. 

I will also remove the logic that user can insert module name as args to only parse certain genesis modules. I assume specifying in the config.yaml would be enough.

rmontagnin: @huichiaotsou I think we can keep the logic of only specifying some modules to be parsed I believe. It's going to be useful if we add new modules on the way. We can however simplify it, I can show you how in the next review after all the other comments are solved

modules/auth/handle_genesis.go
```diff
@@ -5,13 +5,10 @@ import (
 	"fmt"
 
 	tmtypes "github.com/tendermint/tendermint/types"
-
-	"github.com/rs/zerolog/log"
 )
 
 // HandleGenesis implements modules.GenesisModule
 func (m *Module) HandleGenesis(_ *tmtypes.GenesisDoc, appState map[string]json.RawMessage) error {
-	log.Debug().Str("module", "auth").Msg("parsing genesis")
```
rmontagnin, line 14: Can we bring back those logs please? I think they are interested also when parsing the genesis by hand

cmd/parse-genesis/parseGenesis.go
```diff
@@ -0,0 +1,45 @@
...
+
+			genesisDoc, genesisState, err := junoutils.GetGenesisDocAndState(config.Cfg.Parser.GenesisFilePath, parseCtx.Node)
+			if err != nil {
+				return fmt.Errorf("error while getting genesis doc or state: %s", err)
+			}
+
+			invalidInputs, err := utils.ParseGenesis(parseCtx.Modules, genesisDoc, genesisState, args)
+			if err != nil {
+				return fmt.Errorf("error while parsing genesis: %s", err)
+			}
```
rmontagnin: Instead of using this logic, I think a better approach could be the following: 

1. Take the list of modules to parse: if provided by the user, use those. Otherwise use the default list 
2. Iterate through all the specified modules, and find anyone that does not match. For each module name:
   - if found, add that module to a list of modules to parse 
   - if not found, return an error
3. Once we have validated all the modules names, simply go through all the modules to parse and call `HandleGenesis`

All of this can be done inside this method rather than defining another method

huichiaotsou: cool, I have made the changes. As for the module not found, should we return error or just print the log that the module is not registered?

rmontagnin: @huichiaotsou I think it's better to return an error here since it's something that is specified by hand from the user and they should check out whether it was a typo or an error inside the config

rmontagnin (approved):

PR #276, opened 2021-11-26
feat: handle feegrant module

### LINKED ISSUE #278: Add feegrant sdk module

MonikaCat:
## Description

Closes #278 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


database/schema/12-feegrant.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE fee_grant_allowance
+(
+    id                 SERIAL      NOT NULL PRIMARY KEY,
+    grantee            TEXT        NOT NULL REFERENCES account (address),
+    granter            TEXT        NOT NULL REFERENCES account (address),
+    allowance          JSONB       NOT NULL,
+    height             BIGINT      NOT NULL
```
rmontagnin: The `feegrant` module does not allow having more than one allowance per grantee-granter at the same time. For this reason, we should add a `UNIQUE` constraint here: 
```suggestion
    height             BIGINT      NOT NULL,
    CONSTRAINT unique_fee_grant_allowance UNIQUE(grantee, granter) 
```

database/feegrant.go
```diff
@@ -0,0 +1,46 @@
...
+	feegranttypes "github.com/cosmos/cosmos-sdk/x/feegrant"
+	"github.com/forbole/bdjuno/v2/types"
+)
+
+// SaveFeeGrantAllowance allows to store the fee grant allowances for the given block height
+func (db *Db) SaveFeeGrantAllowance(allowance *feegranttypes.MsgGrantAllowance, height int64) error {
+	stmt := `
+INSERT INTO fee_grant_allowance(grantee, granter, allowance, height) 
+VALUES ($1, $2, $3, $4) 
+ON CONFLICT DO NOTHING`
```
rmontagnin: Once the constraint is implemented, we need to change this to make sure that if we're updating multiple heights at the same time, we do not override a grant on a higher height. To do this, we can do as we do in all other tables: 
```suggestion
INSERT INTO fee_grant_allowance(grantee, granter, allowance, height) 
VALUES ($1, $2, $3, $4) 
ON CONFLICT ON CONSTRAINT unique_fee_grant_allowance DO UPDATE 
    SET allowance = excluded.allowance,
        height = excluded.height
WHERE fee_grant_allowance.height <= excluded.height`
```

database/feegrant.go
```diff
@@ -0,0 +1,46 @@
...
+
+	allowanceJSON, err := codec.ProtoMarshalJSON(allowance.Allowance, nil)
+	if err != nil {
+		return fmt.Errorf("error while marshaling grant allowance: %s", err)
+	}
```
huichiaotsou, line 32: just curious here, should we unpack Allowance? I am not sure which one is better for front end, simply storing raw json or unpacked, or maybe it's the same :p

database/feegrant.go
```diff
@@ -0,0 +1,46 @@
...
+
+	// Store the accounts
+	var accounts []types.Account
+	accounts = append(accounts, types.NewAccount(allowance.Grantee), types.NewAccount(allowance.Granter))
+	err = db.SaveAccounts(accounts)
+	if err != nil {
+		return fmt.Errorf("error while storing fee grant allowance accounts: %s", err)
+	}
```
rmontagnin: Can we move this code as the first thing inside the function? It would make the overall code easier to read

database/feegrant.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"fmt"
+
+	"github.com/cosmos/cosmos-sdk/codec"
+	feegranttypes "github.com/cosmos/cosmos-sdk/x/feegrant"
+	"github.com/forbole/bdjuno/v2/types"
+)
+
+// SaveFeeGrantAllowance allows to store the fee grant allowances for the given block height
+func (db *Db) SaveFeeGrantAllowance(allowance *feegranttypes.MsgGrantAllowance, height int64) error {
```
rmontagnin: I think that instead of passing a `MsgGrantAllowance` here we should build a new type that is 

```go
type Grant struct {
   feegrant.Grant
   Heght int64
}
``` 
This way it's easier to incapsualte things here and pass the `SaveFeeGrantAllowance` a single parameter instead of multiple ones

database/feegrant.go
```diff
@@ -0,0 +1,46 @@
...
+	_, err = db.Sql.Exec(stmt, allowance.Grantee, allowance.Granter, allowanceJSON, height)
+	if err != nil {
+		return fmt.Errorf("error while storing fee grant allowance: %s", err)
+	}
+
+	return nil
+}
+
+// RevokeFeeGrantAllowance removes the fee grant allowances data from the database
+func (db *Db) RevokeFeeGrantAllowance(grantee string, granter string) error {
```
rmontagnin: Since BDJuno can be handling multiple blocks at the same time, we also need to make sure we pass the `height` to the delete methods in order to make sure we delete the rows only for older heights, and not for newer ones that might have been parsed at the same time. For this reason, I think we can define a new type that is 

```go
type GrantRemoval struct {
  Grantee string
  Granter string
  Height int64 
}
```

and then we can pass a single `grantDeletion` to the `RevokeFeeGrantAllowance` method.  

Also, this method should be named `DeleteFeeGrant` instead cause it might be used elsewhere (eg. when a feegrant expires or when the user uses all the allowance he had available).

database/feegrant.go
```diff
@@ -0,0 +1,46 @@
...
+	_, err := db.Sql.Exec(`DELETE FROM fee_grant_allowance WHERE grantee = $1 AND granter = $2`, grantee, granter)
```
rmontagnin: For the same reason explained above, we should make sure we remove only older rows: 

```go
    stmt := `DELETE FROM fee_grant_allowance WHERE grantee = $1 AND granter = $2 AND height <= $3`
    _, err := db.Sql.Exec(stmt, grantee, granter, height)
```

modules/feegrant/module.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/cosmos/cosmos-sdk/codec"
+
+	"github.com/forbole/bdjuno/v2/database"
+
+	"github.com/forbole/juno/v2/modules"
+)
+
+var (
+	_ modules.Module        = &Module{}
+	_ modules.MessageModule = &Module{}
```
rmontagnin, line 14: Inside the `feegrant` module, grants can be removed for three main reasons: 
1. The grant is expired
2. The grantee has used all the granted amount 
3. The granter revokes the grant

In all these cases, the module will emit an event with the following attributes: 
- Type: `feegrant.EventTypeRevokeFeeGrant` (`revoke_feegrant`)
- Attributes:
   - `feegrant.AttributeKeyGranter` with value the address of the granter
   - `feegrant.AttributeKeyGrantee` with value the address of the grantee

To properly handle those cases, we should implement the `BlockModule` interface for this method and leverage the `EndBlock` events list to find all events that signal a grant should be removed. For each even found, we should remove that grant from the database.

huichiaotsou (commented):
Since we are adding a new table, should we also keep track of hasura metadata?

database/feegrant.go
rmontagnin: I think storing the allowance as JSON encoded is perfectly fine honestly. The frontend can then get a glimpse of the type of allowance (`BasicAllowance`, `PeriodicAllowance`, `AllowedMsgAllowance`, etc) and render them the best way they think.

modules/feegrant/module.go
MonikaCat: I added the `BlockModule` and handled the EventTypeRevokeFeeGrant event type, but when I test it on my local, the `EndBlock` array is always empty and it doesn't trigger the method to remove expired grant

database/schema/12-feegrant.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE fee_grant_allowance
+(
+    id                 SERIAL      NOT NULL PRIMARY KEY,
+    grantee            TEXT        NOT NULL REFERENCES account (address),
+    granter            TEXT        NOT NULL REFERENCES account (address),
```
rmontagnin: I think we should rename those so that the GraphQL relationships will have better names: 
```sql
    grantee_address            TEXT        NOT NULL REFERENCES account (address),
    granter_address            TEXT        NOT NULL REFERENCES account (address),
```

hasura/metadata/databases/bdjuno/tables/public_fee_grant_allowance.yaml
```diff
@@ -0,0 +1,21 @@
+table:
+  name: fee_grant_allowance
+  schema: public
+object_relationships:
+- name: account
+  using:
+    foreign_key_constraint_on: grantee
+- name: accountGranter
+  using:
+    foreign_key_constraint_on: granter
```
rmontagnin: Once the SQL field are renamed, we can rename those as well
```yaml
object_relationships:
- name: grantee
  using:
    foreign_key_constraint_on: grantee_address
- name: granter
  using:
    foreign_key_constraint_on: granter_address
```

modules/feegrant/handle_block.go
```diff
@@ -0,0 +1,64 @@
...
+			return fmt.Errorf("error while getting fee grant grantee address: %s", err)
+		}
+		height, err := juno.FindAttributeByKey(event, "height")
+		if err != nil {
+			return fmt.Errorf("error while getting fee grant height: %s", err)
+		}
+		blockHeight, err := strconv.ParseInt(height.String(), 10, 64)
+		if err != nil {
+			return fmt.Errorf("error while parsing blockHeight: %s", err)
+		}
```
rmontagnin: No need to do this, we can leverage the `height` param passed to `removeExpireFreeGrantAllowance` directly

types/feegrant.go
```diff
@@ -0,0 +1,32 @@
+package types
+
+import feegranttypes "github.com/cosmos/cosmos-sdk/x/feegrant"
+
+// FeeGrant represents the x/feegrant module
+type FeeGrant struct {
+	*feegranttypes.MsgGrantAllowance
```
rmontagnin: I think that instead of having a `MsgGrantAllowance` here it would be better to use the correct type that identifies a grant: `feegrant.Grant`. This would make sure we are dealing with the same types used by the Cosmos SDK to store the data

utils/node.go
```diff
@@ -27,3 +27,12 @@ func QueryTxs(node node.Node, query string) ([]*coretypes.ResultTx, error) {
 
 	return txs, nil
 }
+
+// QueryBlock queries block details from the given node
+func QueryBlock(node node.Node, blockHeight int64) (*coretypes.ResultBlock, error) {
```
rmontagnin: Is this method really necessary?

MonikaCat: Yes this method helps to get the block details for the `feegrant` fix command. We also reuse it in PR #248 to get the block results for `block` fix command

modules/feegrant/module.go
rmontagnin: @MonikaCat I've further investigated into this and I found out it's caused by a flawless logic inside the `feegrant` module of the Cosmos SDK. I've now opened an issue here: https://github.com/cosmos/cosmos-sdk/issues/10704. I think we can keep it like you implemented it right now (using the `EndBlocker`) and then simply fix it once that's solved

Makefile
```diff
@@ -52,7 +52,7 @@ stop-docker-test:
 
 start-docker-test: stop-docker-test
 	@echo "Starting Docker container..."
-	@docker run --name bdjuno-test-db -e POSTGRES_USER=bdjuno -e POSTGRES_PASSWORD=password -e POSTGRES_DB=bdjuno -d -p 5433:5432 postgres
+	@docker run --name bdjuno-test-db -e POSTGRES_USER=bdjuno -e POSTGRES_PASSWORD=password -e POSTGRES_DB=bdjuno -d -p 6433:5432 postgres
```
rmontagnin, line 55: This it to be able to use the `make start-docker-test` even if you have multiple PostgreSQL instances running

database/database_test.go
```diff
@@ -48,7 +48,7 @@ func (suite *DbTestSuite) SetupTest() {
 	dbCfg := dbconfig.NewDatabaseConfig(
 		"bdjuno",
 		"localhost",
-		5433,
+		6433,
```
rmontagnin, line 51: Same as above

modules/feegrant/handle_block.go
```diff
@@ -0,0 +1,55 @@
...
+		granterAddress, err := juno.FindAttributeByKey(event, feegranttypes.AttributeKeyGranter)
+		if err != nil {
+			return fmt.Errorf("error while getting fee grant granter address: %s", err)
+		}
+		granteeAddress, err := juno.FindAttributeByKey(event, feegranttypes.AttributeKeyGrantee)
+		if err != nil {
+			return fmt.Errorf("error while getting fee grant grantee address: %s", err)
+		}
+		allowanceToRemove := types.NewGrantRemoval(granteeAddress.String(), granterAddress.String(), height)
+		err = m.db.DeleteFeeGrantAllowance(allowanceToRemove)
```
rmontagnin: If we use `granteeAddress.String()` here we would end up having a stringified ABCI event. Instead, we need to get the `value` if that event. Also we can merge all in one line: 
```suggestion
		err = m.db.DeleteFeeGrantAllowance(types.NewGrantRemoval(string(granteeAddress.Value), string(granterAddress.Value), height))
```

modules/feegrant/handle_block.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/rs/zerolog/log"
+	tmctypes "github.com/tendermint/tendermint/rpc/core/types"
+
+	"github.com/forbole/bdjuno/v2/types"
+)
+
+// HandleBlock implements BlockModule
+func (m *Module) HandleBlock(
+	block *tmctypes.ResultBlock, res *tmctypes.ResultBlockResults, _ []*juno.Tx, vals *tmctypes.ResultValidators,
+) error {
```
rmontagnin: We don't use the `vals` so we can mark it as `_`: 

```suggestion
func (m *Module) HandleBlock(
	block *tmctypes.ResultBlock, res *tmctypes.ResultBlockResults, _ []*juno.Tx, _ *tmctypes.ResultValidators,
) error {
```

modules/feegrant/handle_msg.go
```diff
@@ -0,0 +1,47 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	feegranttypes "github.com/cosmos/cosmos-sdk/x/feegrant"
+	juno "github.com/forbole/juno/v2/types"
+
+	"github.com/forbole/bdjuno/v2/types"
+)
+
+// HandleMsg implements modules.MessageModule
+func (m *Module) HandleMsg(index int, msg sdk.Msg, tx *juno.Tx) error {
```
rmontagnin: The `index` is not used so we can mark it `_`
```suggestion
func (m *Module) HandleMsg(_ int, msg sdk.Msg, tx *juno.Tx) error {
```

modules/feegrant/handle_msg.go
```diff
@@ -0,0 +1,47 @@
...
+	return nil
+}
+
+// HandleMsgGrantAllowance allows to properly handle a MsgGrantAllowance
+func (m *Module) HandleMsgGrantAllowance(tx *juno.Tx, msg *feegranttypes.MsgGrantAllowance) error {
+	allowance, err := msg.GetFeeAllowanceI()
+	if err != nil {
+		return fmt.Errorf("error while getting fee allowance: %s", err)
+	}
+	feeGrant, err := feegranttypes.NewGrant(sdk.AccAddress(msg.Granter), sdk.AccAddress(msg.Grantee), allowance)
```
rmontagnin: Instead of using `sdk.AccAddress(address)` we should use `sdk.AccAddressFromBech32` instead:
```suggestion
	granter, err := sdk.AccAddressFromBech32(msg.Granter)
	if err != nil {
		return fmt.Errorf("error while parsing granter address: %s", err)
	}
	grantee, err := sdk.AccAddressFromBech32(msg.Grantee)
	if err != nil {
		return fmt.Errorf("error while parsing grantee address: %s", err)
	}
	feeGrant, err := feegranttypes.NewGrant(granter, grantee, allowance)
```

modules/feegrant/handle_msg.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return fmt.Errorf("error while getting new grant allowance: %s", err)
+	}
+	feeGrantAllowance := types.NewFeeGrant(feeGrant, tx.Height)
+	return m.db.SaveFeeGrantAllowance(feeGrantAllowance)
```
rmontagnin:
```suggestion
	return m.db.SaveFeeGrantAllowance(types.NewFeeGrant(feeGrant, tx.Height))
```

modules/feegrant/handle_msg.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// HandleMsgRevokeAllowance allows to properly handle a MsgRevokeAllowance
+func (m *Module) HandleMsgRevokeAllowance(tx *juno.Tx, msg *feegranttypes.MsgRevokeAllowance) error {
+	allowanceToDelete := types.NewGrantRemoval(msg.Grantee, msg.Granter, tx.Height)
+	return m.db.DeleteFeeGrantAllowance(allowanceToDelete)
```
rmontagnin:
```suggestion
	return m.db.DeleteFeeGrantAllowance(types.NewGrantRemoval(msg.Grantee, msg.Granter, tx.Height))
```

modules/feegrant/module.go
```diff
@@ -0,0 +1,37 @@
...
+type Module struct {
+	cdc codec.Codec
+	db  *database.Db
+}
+
+// NewModule returns a new Module instance
+func NewModule(
+	cdc codec.Codec,
+	db *database.Db,
+) *Module {
```
rmontagnin:
```suggestion
func NewModule(cdc codec.Codec, db *database.Db) *Module {
```

modules/feegrant/module.go
MonikaCat: Ahh okay! In that case let's merge it when you approve the changes 👍🏻

cmd/fix/feegrant/allowance.go
```diff
@@ -0,0 +1,100 @@
...
+			// Get latest height
+			height, err := parseCtx.Node.LatestHeight()
+			if err != nil {
+				return fmt.Errorf("error while getting chain latest block height: %s", err)
+			}
+
+			startHeight := config.Cfg.Parser.StartHeight
+
+			// Handle messages realated to feegrant allowance of each block from the start height
+			for h := startHeight; h <= height; h++ {
```
rmontagnin: Instead of parsing all the blocks from the starting height till the latest height, there is a much more efficient way of fixing past fee grant-related transactions: 
1. Get all past transactions that contained a `MsgGrantAllowance` or `MsgRevokeAllowance`
2. Order the transactions based on the ascending height 
3. Parse the transactions messages

You can see something similar [here](https://github.com/desmos-labs/djuno/blob/main/cmd/fix/profiles/cmd.go#L37)

MonikaCat: Ok! I have updated the code and tested it locally

cmd/fix/feegrant/allowance.go
```diff
@@ -0,0 +1,113 @@
...
+							continue
+						}
+
+						err = feegrantModule.HandleMsg(index, msg, transaction)
+						if err != nil {
+							return fmt.Errorf("error while handling feegrant module message: %s", err)
+						}
+					}
+				}
+			}
```
rmontagnin: Instead of searching for all given and revoked grants for all accounts there is a way to search for all grant and revoke grant transactions. This should allow us to both remove the `for` cycle as well as removing a lot of code here: 

```suggestion
			// Collect all the transactions
			var txs []*tmctypes.ResultTx

			// Get all the MsgGrantAllowance txs
			query := fmt.Sprintf("message.action='%s'", feegranttypes.EventTypeSetFeeGrant)
			grantAllowanceTxs, err := utils.QueryTxs(parseCtx.Node, query)
			if err != nil {
				return err
			}
			txs = append(txs, grantAllowanceTxs...)

			// Get all the MsgRevokeAllowance txs
			query = fmt.Sprintf("message.action='%s'", feegranttypes.EventTypeRevokeFeeGrant)
			revokeAllowanceTxs, err := utils.QueryTxs(parseCtx.Node, query)
			if err != nil {
				return err
			}
			txs = append(txs, revokeAllowanceTxs...)

			// Sort the txs based on their ascending height
			sort.Slice(txs, func(i, j int) bool {
				return txs[i].Height < txs[j].Height
			})

			for _, tx := range txs {
				log.Debug().Int64("height", tx.Height).Msg("parsing transaction")
				transaction, err := parseCtx.Node.Tx(hex.EncodeToString(tx.Tx.Hash()))
				if err != nil {
					return err
				}

				// Handle only the MsgGrantAllowance and MsgRevokeAllowance instances
				for index, msg := range transaction.GetMsgs() {
					_, isMsgGrantAllowance := msg.(*feegranttypes.MsgGrantAllowance)
					_, isMsgRevokeAllowance := msg.(*feegranttypes.MsgRevokeAllowance)

					if !isMsgGrantAllowance && !isMsgRevokeAllowance {
						continue
					}

					err = feegrantModule.HandleMsg(index, msg, transaction)
					if err != nil {
						return fmt.Errorf("error while handling feegrant module message: %s", err)
					}
				}
			}
		}
```

utils/node.go
rmontagnin: This can be deleted now since it's no longer used, right?

rmontagnin (approved):

rmontagnin:
LGTM. As soon as @maiquanghiep accepts this PR it will be merged automatically

huichiaotsou (approved):

PR #353, opened 2022-03-09
feat: updated history module

### LINKED ISSUE #348: Enable Token Price History without History Module
Issue:
Currently token price history is only enabled if `history` module is enabled.

There will be networks that don't require history and I would like to retrieve their token price history in order to display a 7 day chart.

Implementation:
Separate Token Price History from history. Log the price once a day (seriously once a day should be enough)

MonikaCat:
## Description

Closes: #348 

- Updated periodic ops to query token price and store it inside `token_price` table every 2 mins
- Added periodic ops to history module to query historic token price and store it inside `token_price_history` table every 1hr 
---

### Author Checklist

I have...

### Reviewers Checklist

I have...


rmontagnin:
What is the current `history` module storing since we replaced all historic data with Hasura actions? Cause I think it might be easier to move the historic token storage inside the `pricefeed` module itself so one module does everything there's needed rather than splitting into another module (which needs to be enabled)

MonikaCat:
Currently history module is storing historical price data only. Sure, I moved it inside `pricefeed` module as per you suggestion 👍🏻

modules/pricefeed/handle_periodic_operations.go
```diff
@@ -14,13 +15,20 @@ import (
 func (m *Module) RegisterPeriodicOperations(scheduler *gocron.Scheduler) error {
 	log.Debug().Str("module", "pricefeed").Msg("setting up periodic tasks")
 
-	// Fetch total supply of token in 30 seconds each
-	if _, err := scheduler.Every(30).Second().Do(func() {
+	// Fetch total supply of tokens every 2 mins
```
rmontagnin:
```suggestion
	// Fetch the token prices every 2 mins
```

modules/pricefeed/handle_periodic_operations.go
```diff
@@ -14,13 +15,20 @@ import (
...
+	if _, err := scheduler.Every(2).Minutes().Do(func() {
 		utils.WatchMethod(m.updatePrice)
 	}); err != nil {
 		return fmt.Errorf("error while setting up pricefeed period operations: %s", err)
 	}
 
+	// Fetch total supply of tokens every 1hr to store historical price data
```
rmontagnin:
```suggestion
	// Update the historical token prices every 1 hour
```

modules/pricefeed/handle_periodic_operations.go
```diff
@@ -54,5 +73,27 @@ func (m *Module) updatePrice() error {
...
+
+// updatePricesHistory fetches total amount of coins in the system from RPC
+// and stores historical perice data inside the database
+func (m *Module) updatePricesHistory() error {
+	log.Debug().
+		Str("module", "pricefeed").
+		Str("operation", "pricefeed").
+		Msg("updating token price and market cap history")
+
+	prices, err := m.getTokenPrices()
```
rmontagnin, line 89: I removed the part that forced all the prices to have `LastUpdated` set to `time.Now()` as I think it wasn't correct. @MonikaCat Was there a specific reason that one was there?

MonikaCat: @RiccardoM `Last Updated` timestamp is fetched from the coingecko and it reflects the time when the price was last updated. We need the exact value at exact time every 1 hour to plot the graph. If price hasn't change within that 1 hour, the returned timestamp will be the same as one hour ago and it will not be stored in db as it will be a duplicated value. So that's why we should pass the current timestamp rather than the timestamp fetched from coingecko.

rmontagnin: Ok thanks, added back with a comment

rmontagnin (approved):

huichiaotsou (approved):

PR #381, opened 2022-04-08
feat: move Hasura actions to be a module

### LINKED ISSUE #375: merge `hasura-actions` with `bdjuno start` cmd
- Merge hasura-actions to run with BDJuno on `start` command, but keep the ability to run hasura-actions as a separate process
- Add additional section inside config.yaml file to use when running hasura-actions eg.
```
hasura-actions:
    enable: true
    rpc:
            address: http://localhost:26657
            max_connections: 30
        grpc:
            address: http://localhost:9090
            insecure: true
 ```

### PR DESCRIPTION
**Pull request by rmontagnin**
## Description
This PR makes the Hasura actions work just like any other module. The configuration of the module allows to specify another node to get the data from. It also changes how the `migrate v3` command works by adding the configuration for the new `actions` module.

Closes: #375

---

### Author Checklist

I have...

### Reviewers Checklist

I have...

### CODE REVIEW

File: modules/actions/module.go [hunk 1]
```diff
@@ -0,0 +1,61 @@
...
+	node    node.Node
+	sources *modulestypes.Sources
+}
+
+func NewModule(cfg config.Config, encodingConfig *params.EncodingConfig) *Module {
+	actionsCfg, err := ParseConfig(cfg.GetBytes())
+	if err != nil {
+		panic(err)
+	}
+
```
huichiaotsou, line 34: do we want to hint users if `actions.port` is not set in `config.yaml` or it's unnecessary? for example:

```go
	if actionsCfg == nil {
		panic("hasura actions port needs to be specified in config.yaml")
	}
```

huichiaotsou: do we want to hint users if `actions.port` is not set in `config.yaml` or it's unnecessary? for example:

```go
	if actionsCfg == nil {
		panic("hasura actions port needs to be specified in config.yaml")
	}
```

huichiaotsou: or maybe we want to add `actions.port` to `migrate v3`'s migrate config  ?

MonikaCat:
I think we should also add `actions` to config file on `init` command and add `node` section for rpc/grpc address in default config file, what do you think?

modules/actions/module.go
rmontagnin: The addition of the `actions` section (as well as the `actions` module inside the list of modules) has already been added to the v3 migration here: https://github.com/forbole/bdjuno/pull/381/files#diff-067c77996d22ddcb35de335fd771b1eafc125879a0ab51b91ce10bf42436d2a1R61

rmontagnin:
> I think we should also add `actions` to config file on `init` command and add `node` section for rpc/grpc address in default config file, what do you think?

Added, thanks!

modules/actions/config.go
```diff
@@ -19,6 +19,14 @@ func NewConfig(port uint, remoteDetails *remote.Details) *Config {
 	}
 }
 
+// DefaultConfig returns the default configuration
+func DefaultConfig() Config {
+	return Config{
+		Port: 3000,
+		Node: nil,
```
MonikaCat, line 26:
```suggestion 
		Node: remote.DefaultDetails(),
```
We can pass default node config here

cmd/migrate/v3/migrate.go
```diff
@@ -0,0 +1,66 @@
...
+		return Config{}, fmt.Errorf("error while reading v2 config: %s", err)
+	}
+
+	// Enable the actions module if not enabled
+	if !cfg.Chain.IsModuleEnabled(actions.ModuleName) {
+		cfg.Chain.Modules = append(cfg.Chain.Modules, actions.ModuleName)
+	}
+
+	if cfg.Actions == nil {
+		cfg.Actions = actions.NewConfig(3000, nil)
```
MonikaCat, line 62:
```suggestion
		cfg.Actions = actions.NewConfig(3000, remote.DefaultDetails())
```
Also here

modules/actions/config.go
rmontagnin: I've actually intentionally left it `nil` there because I think that the default behavior should be to use the same node that other data is parsed from when handling actions. Using a custom node for actions should be something optional and not required. Do you think it's better to force the users to specify another node there by default instead?

MonikaCat: Alright let's keep it this way and I will update the docs to explain how to setup custom node address for hasura actions

MonikaCat:
I am testing it on my local and I can't get the hasura actions to work properly. I am getting `internal server error` inside hasura console and in docker logs this error pops up `Spock Error while handling ["v1","graphql"]: InvalidUrlException "host.docker.internal:3000/account_balance" "Invalid scheme"` 🤔

rmontagnin:
> I am testing it on my local and I can't get the hasura actions to work properly. I am getting `internal server error` inside hasura console and in docker logs this error pops up `Spock Error while handling ["v1","graphql"]: InvalidUrlException "host.docker.internal:3000/account_balance" "Invalid scheme"` thinking

Does anything else work if you don't use Hasura actions? What's the config?

MonikaCat:
Yes bdjuno works, I migrated to v3 config file and just set port to 3000 inside actions section like so:
```
actions:
    port: 3000
```

MonikaCat:
After investigating with @RiccardoM we've found out that this error was caused by missing `http://` prefix inside docker' `ACTION_BASE_URL` flag

MonikaCat (approved):
LGTM

huichiaotsou (approved):

PR #403, opened 2022-05-09
fix: filter empty validator address while storing validator signing info

huichiaotsou:
## Description

Closes: #XXXX

While querying for slashing info, the node would return empty address and it would result in error while updating signing infos in the DB: 
```shell
5:40PM ERR error while handling block err="error while updating signing info: 
error while storing validators signing infos: pq: ON CONFLICT DO UPDATE command 
cannot affect row a second time" height=5200791 module=slashing
```

---

### Author Checklist

I have...

- [ ] targeted the correct branch

### Reviewers Checklist

I have...


MonikaCat (approved):
LGTM

database/slashing.go
```diff
@@ -18,15 +18,19 @@ INSERT INTO validator_signing_info
     (validator_address, start_height, index_offset, jailed_until, tombstoned, missed_blocks_counter, height)
 VALUES `
 	var args []interface{}
+	var counter int
 
-	for i, info := range infos {
-		ii := i * 7
+	for _, info := range infos {
+		if info.ValidatorAddress != "" {
```
rmontagnin: Instead of wrapping all the logic inside this `if`, I would prefer if we do the other way around: we check if the condition is valid, and we continue the loop if that's the case (skipping the iteration): 

```go
for _, info := range infos {
  if info.ValidatorAddress == "" {
    continue
  }
  
  // Insert the data here
}
```

This would end up in a much more readable code overall

rmontagnin (approved):

PR #411, opened 2022-05-26
fix: query db instead for validator tombstoned status

huichiaotsou:
## Description

Closes: #XXXX

currently we are encountering a lot of errors while getting validator tombstoned status in the staking module. 
``` 
error="error while getting validator signing info: 
rpc error: code = Unavailable desc = transport is closing" height=621512 module=staking
```

This PR replaces querying the slashing info to the node with querying to the database validator_signing_info table. 

---

### Author Checklist

I have...

### Reviewers Checklist

I have...


database/slashing.go
```diff
@@ -71,3 +71,20 @@ WHERE slashing_params.height <= excluded.height`
 
 	return nil
 }
+
+// GetValidatorTombstoneStatus gets the validator latest tombstoned status from the database
```
MonikaCat:
```suggestion
// GetValidatorTombstonedStatus gets the validator latest tombstoned status from the database
```

MonikaCat (approved):
tested, LGTM

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 #### Bank module
 - ([\#410](https://github.com/forbole/bdjuno/pull/410)) Change total supply query from only 1 page to all pages
 
+#### Staking module
+- ([\#411](https://github.com/forbole/bdjuno/pull/411)) Get validator tombstoned status from the database instead of querying to the node
```
rmontagnin: Can we move this to a new `Unreleased` section above the `Version v3.1.0` one?

database/slashing.go
```diff
@@ -71,3 +71,20 @@ WHERE slashing_params.height <= excluded.height`
 
 	return nil
 }
+
+// GetValidatorTombstonedStatus gets the validator latest tombstoned status from the database
+func (db *Db) GetValidatorTombstonedStatus(valConsAddress string) (bool, error) {
```
rmontagnin, line 76: I only have question, which I don't remember the answer to: if we're querying this value, when are we updating it? What would happen is a validator is jailed or tombstoned in a new block? Is the value updated somewhere else or what?

huichiaotsou: hey Ricky it's just wingman requested to have the `tombstone` status in the `validator_status`(x/staking) table while we used to save it only in `validator_signing_info`(x/slashing). So:
- the `tombstone` status is updated with the slashing module
- in the staking module the status is grabbed from `validator_signing_info`  and stored again in `validator_status`

rmontagnin: @ryuash Can we have some more context on why this is needed/required/preferable?

huichiaotsou: Otherwise we just get the `tombstone status` from the `validator_signing_info` table, gql query be something like this
```gql
query MyQuery {
  validator_status {
    height
    jailed
    status
    validator_address
    validator {
      validator_signing_infos {
        tombstoned
      }
    }
  }
}
```

in this case I will remove the tomstoned column from `validator_status` table.

would this be reasonable for you @yayay927 ?

ryuash: I remember this was requested months back. I no longer remember the purpose tbh

huichiaotsou: then I think we can remove this part of codes because it keeps causing "transport is closing" error while calling the node :-/ we should then alternatively JOIN the `validator_signing_info` table for `tombstone` status.

ryuash: OH SORRY I REMEMBER NOW. Yes this was requested during the first wave of tombstones validators as we had no way of seeing who was tombstoned in bdjuno at the time. 

It's currently being used in the frontend so a migration would be needed if you're making any changes.

Would moving it to a different table be better or? I figured bdjuno was listening for the tombstone event to update this but I could be wrong

huichiaotsou: Cool so front end will get the tombstone from validator signing info first and then we can remove those codes from BDjuno safely :D

rmontagnin: Honestly if BD is fine as it is right now, I would avoid updating anything

huichiaotsou: I see, but we actually get a lot of errors like this while running staking module
```
error="error while getting validator signing info: 
rpc error: code = Unavailable desc = transport is closing" height=621512 module=staking
```

saving tombstone twice is weird no? when we can just join the tables

PR #652, opened 2023-09-20
feat: update `x/gov` module parsing

MonikaCat:
## Description

Closes: #XXXX
[BDU-11](https://forbole.atlassian.net/browse/BDU-11)
[BDU-771](https://forbole.atlassian.net/browse/BDU-771)
[BDU-1077](https://forbole.atlassian.net/browse/BDU-1077)

Changes:
`gov` module:
- Added handler on every block to check and update proposal status if `active_proposal` event has been emitted inside the EndBlockEvents
- Added periodic operations to update `proposal staking pool snapshots` every 5 mins 
- Added periodic operations to update `proposal tally results` every 5 mins 
- Added handler to update `tally results` for given proposal on every `MsgVote` and `MsgVoteWeighted` msgs
- Added handler for rpc error `rpc error returning code = NotFound desc = proposal x doesn't exist` returned from node  sometimes when processing `MsgSubmitProposal` msg and querying the node for proposal details
- Added amount to `unique_deposit` constraint to allow to store different deposits from the same address for the same proposal 
- Added `weight` column to `proposal_vote` table and added `option` to `unique_vote` constraint
- Added handler for `MsgVoteWeighted` msgs to correctly store votes in db
- Removed `auth` module from expected modules

`staking` module:
- Added periodic operations to update validators statuses, voting power and proposal validators status snapshots every 5 mins 
- Added handler to update validators statuses, voting powers and proposals validators status snapshots when there is a VP change on `MsgDelegate`, `MsgBeginRedelegate` and `MsgUndelegate` msgs

---

### Author Checklist

I have...

- [x] updated the relevant documentation or specification

### Reviewers Checklist

I have...

[BDU-11]: https://forbole.atlassian.net/browse/BDU-11?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BDU-771]: https://forbole.atlassian.net/browse/BDU-771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BDU-1077]: https://forbole.atlassian.net/browse/BDU-1077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/gov/utils_proposal.go
```diff
@@ -43,40 +44,67 @@ func (m *Module) UpdateProposal(height int64, blockTime time.Time, id uint64) er
...
+			proposal.VotingEndTime,
+		),
+	)
+}
+
+// UpdateProposalsStakingPoolSnapshot updates
+// staking pool snapshots for active proposals
+func (m *Module) UpdateProposalsStakingPoolSnapshot() error {
+	log.Debug().Str("module", "gov").Msg("refreshing proposal staking pool snapshots")
+	blockTime, err := m.db.GetLastBlockTimestamp()
```
zura35: I see that the queries to get last block ts and height are the same, only by returning different columns; given that getting the block height needs not depend on the block time, wonder if we can simply fetch both columns in the same db query to reduce roundtrip as much as possible?

modules/gov/utils_proposal.go
```diff
@@ -10,22 +10,23 @@ import (
...
 
-	distrtypes "github.com/cosmos/cosmos-sdk/x/distribution/types"
-
 	gov "github.com/cosmos/cosmos-sdk/x/gov/types"
 )
 
-func (m *Module) UpdateProposal(height int64, blockTime time.Time, id uint64) error {
+// UpdateProposalStatus queries the latest details of given proposal ID, updates it's status
+// in database and handles changes if the proposal has been passed.
+func (m *Module) UpdateProposalStatus(height int64, blockTime time.Time, id uint64) error {
```
zura35: prolly can remove unused blockTime

modules/gov/utils_proposal.go
```diff
@@ -43,40 +44,67 @@ func (m *Module) UpdateProposal(height int64, blockTime time.Time, id uint64) er
...
 	}
 
-	err = m.updateProposalTallyResult(proposal)
+	err = m.handlePassedProposal(proposal, height)
 	if err != nil {
-		return fmt.Errorf("error while updating proposal tally result: %s", err)
+		return fmt.Errorf("error while handling passed proposals: %s", err)
 	}
 
-	err = m.updateAccounts(proposal)
```
zura35, line 51: May I know what is the purpose of removing this? :o

database/auth.go
```diff
@@ -98,8 +98,14 @@ func (db *Db) storeVestingAccount(account exported.VestingAccount) (int, error)
 			start_time = excluded.start_time
 			RETURNING id `
 
+	// Store the vesting account
+	err := db.SaveAccounts([]types.Account{types.NewAccount(account.GetAddress().String())})
```
zura35, line 102: I see that `SaveVestingAccounts()` is called in `modules/auth/handle_genesis.go` as well where `SaveAccounts()` will also be called in the same block. I wonder if this line would better be moved to `cmd/parse/auth/vesting.go` if your intention is to also insert into `account` table for that parse command?

database/gov.go
```diff
@@ -220,27 +230,28 @@ func (db *Db) SaveDeposits(deposits []types.Deposit) error {
...
 	err := db.SaveAccounts(accounts)
 	if err != nil {
-		return fmt.Errorf("error while storing depositor account: %s", err)
+		return fmt.Errorf("error while storing depositors accounts: %s", err)
 	}
 
 	query = query[:len(query)-1] // Remove trailing ","
 	query += `
 ON CONFLICT ON CONSTRAINT unique_deposit DO UPDATE
-	SET amount = excluded.amount,
```
zura35: Why do we ignore amount update? :o

database/schema/08-gov.sql
```diff
@@ -29,7 +29,7 @@ CREATE TABLE proposal_deposit
     amount            COIN[],
     timestamp         TIMESTAMP,
     height            BIGINT  NOT NULL,
-    CONSTRAINT unique_deposit UNIQUE (proposal_id, depositor_address)
+    CONSTRAINT unique_deposit UNIQUE (proposal_id, depositor_address, amount)
```
zura35: Does it mean in a proposal it is indeed not possible to have multiple deposits with the same amount from the same depositor address? Or this change is simply taking advantage of indexing? @@

zura35 (commented):
Cuz I'm not familiar with bdjuno enough, you will see questions more than comments 😂 And wonder if there are anyone else to approve the changes but if not I can help approve if needed 🙏

database/schema/08-gov.sql
MonikaCat: I added this because sometimes the same address will pay the deposit in more than one txs (eg first deposit 100 DARIC, and later deposit 400 DARIC in a different tx) and right now only the latest deposit will be stored for given address inside the `proposal_deposit` table so by updating this constraint we can store all deposits correctly

MonikaCat: Also, I just updated it and added timestamp to the `unique_deposit` constraint to handle the edge cases where address can pay the same amount of deposit twice in different txs (eg. deposit 100 DARIC first, later deposit 100 DARIC again)

database/gov.go
MonikaCat: The amount here is ignored because with the current flow we should store all proposals deposits, instead of storing just the latest one, hence we shouldn't update the amount value here

modules/gov/utils_proposal.go
MonikaCat: This function was used to store the recipient address from `MsgCommunityPoolSpend` and `MsgExecLegacyContent` msgs, I moved it to `handleMsgSubmitProposal` handler, so we only need to process it once when the proposal is submitted instead of processing it on every block

database/auth.go
MonikaCat: Those two functions `SaveAccounts()` and `SaveVestingAccounts()` are only called when processing the genesis file and we want to first store all base account addresses inside `account` table because the `address` column inside the `vesting_address` table has a reference to `account` table.  I added this line `err := db.SaveAccounts([]types.Account{types.NewAccount(account.GetAddress().String())})` to ensure the vesting address is stored inside `account` table and avoid errors

database/schema/08-gov.sql
zura35: In this case can we also store the `txHash` and use it as a unique constraint? 🤔 It sounds more like a unique identifier for a proposal deposit instead of relying on timestamp and amount?

Or if txn timestamp must be unique across deposits and txns, we can remove `amount` from the unique constraint?

zura35 (commented):
(not change request) I see that we actually use UPSERT (i.e. INSERT ON CONFLICT UPDATE) quite intensively in this project. In general if we are certain about an operation is going to be an UPDATE, it's always better to use `UPDATE` statement directly so psql will not bother attempting to insert and receive an error @@

database/schema/08-gov.sql
MonikaCat: Ok I added `transaction_hash` to `proposal_deposit` table and updated `unique_deposit` constraint. I didn't create the reference to message table because genesis deposits doesn't contain tx hash details, so proposal deposits stored from genesis file will have empty string inside `transaction_hash` column

zura35 (approved):

PR #725, opened 2024-03-21
fix: re-add support for gov v1beta1 messages

dadamu:
## Description

Closes: #XXXX
Currently, gov module is updated to gov v1, however, gov v1beta1 is still alive to chains, Cosmos-SDK doesn't completely remove v1beta1 support. As a result, we should re-add gov v1beta1 support to make sure our proposals data are also updated when gov v1beta1 messages are performed.

Say, some of users on Cheqd is using gov.v1beta1.MsgVote to vote on the proposal.
https://bigdipper.live/cheqd/transactions/19E4F068FBFBC46DA9BEF3E0F7CA908317D5582CAB2A17683138A5248ED34E3C

---

### Author Checklist

I have...

- [ ] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
- [ ] targeted the correct branch

### Reviewers Checklist

I have...

### DIFF
--- a/cmd/parse/gov/proposal.go
+++ b/cmd/parse/gov/proposal.go
@@ -10,6 +10,7 @@ import (
 	modulestypes "github.com/forbole/callisto/v4/modules/types"
 
 	govtypesv1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
+	govtypesv1beta1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1beta1"
 	parsecmdtypes "github.com/forbole/juno/v5/cmd/parse/types"
 	"github.com/forbole/juno/v5/parser"
 	"github.com/forbole/juno/v5/types/config"
@@ -195,29 +196,39 @@ func refreshProposalVotes(parseCtx *parser.Context, proposalID uint64, govModule
 
 		// Handle the MsgVote messages
 		for index, msg := range junoTx.GetMsgs() {
-			_, isMsgVoteV1 := msg.(*govtypesv1.MsgVote)
-			_, isMsgVoteV1Beta1 := msg.(*govtypesv1beta1.MsgVote)
-			_, isMsgVoteWeightedV1 := msg.(*govtypesv1.MsgVoteWeighted)
-			_, isMsgVoteWeightedV1Beta1 := msg.(*govtypesv1beta1.MsgVoteWeighted)
+			var msgProposalID uint64
+
+			switch cosmosMsg := msg.(type) {
+			case *govtypesv1.MsgVote:
+				msgProposalID = cosmosMsg.ProposalId
+
+			case *govtypesv1beta1.MsgVote:
+				msgProposalID = cosmosMsg.ProposalId
+
+			case *govtypesv1.MsgVoteWeighted:
+				msgProposalID = cosmosMsg.ProposalId
+
+			case *govtypesv1beta1.MsgVoteWeighted:
+				msgProposalID = cosmosMsg.ProposalId
 
 			// Skip if the message is not a vote message
-			if !isMsgVoteV1 && !isMsgVoteV1Beta1 && !isMsgVoteWeightedV1 && !isMsgVoteWeightedV1Beta1 {
+			default:
 				continue
-			} else {
-				// check if requested proposal ID is the same as proposal ID returned
-				// from the msg as some txs may contain multiple MsgVote msgs
-				// for different proposals which can cause error if one of the proposals
-				// info is not stored in database
-				if proposalID == msgVote.ProposalId {
-					err = govModule.HandleMsg(index, msg, junoTx)
-					if err != nil {
-						return fmt.Errorf("error while handling MsgVote: %s", err)
-					}
-				} else {
-					// skip votes for proposals with IDs
-					// different than requested in the query
-					continue
+			}
+
+			// check if requested proposal ID is the same as proposal ID returned
+			// from the msg as some txs may contain multiple MsgVote msgs
+			// for different proposals which can cause error if one of the proposals
+			// info is not stored in database
+			if proposalID == msgProposalID {
+				err = govModule.HandleMsg(index, msg, junoTx)
+				if err != nil {
+					return fmt.Errorf("error while handling MsgVote: %s", err)
 				}
+			} else {
+				// skip votes for proposals with IDs
+				// different than requested in the query
+				continue
 			}
 		}
 	}
--- a/cmd/parse/gov/proposal.go
+++ b/cmd/parse/gov/proposal.go
@@ -125,17 +125,13 @@ func refreshProposalDetails(parseCtx *parser.Context, proposalID uint64, govModu
 
 	// Handle the MsgSubmitProposal messages
 	for index, msg := range tx.GetMsgs() {
-		_, isMsgSubmitProposalV1 := msg.(*govtypesv1.MsgSubmitProposal)
-		_, isMsgSubmitProposalV1Beta1 := msg.(*govtypesv1beta1.MsgSubmitProposal)
 
-		// Skip if the message is not a submit proposal message
-		if !isMsgSubmitProposalV1 && !isMsgSubmitProposalV1Beta1 {
-			continue
-		}
-
-		err = govModule.HandleMsg(index, msg, tx)
-		if err != nil {
-			return fmt.Errorf("error while handling MsgSubmitProposal: %s", err)
+		switch msg.(type) {
+		case *govtypesv1.MsgSubmitProposal, *govtypesv1beta1.MsgSubmitProposal:
+			err = govModule.HandleMsg(index, msg, tx)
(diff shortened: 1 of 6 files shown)


modules/gov/handle_msg.go
```diff
@@ -108,39 +104,44 @@ func (m *Module) handleMsgSubmitProposal(tx *juno.Tx, index int, msg *govtypesv1
 		return fmt.Errorf("error while storing proposal recipient: %s", err)
 	}
 
+	// Unpack the proposal interfaces
+	err = proposal.UnpackInterfaces(m.cdc)
+	if err != nil {
+		return fmt.Errorf("error while unpacking proposal interfaces: %s", err)
+	}
```
dadamu, line 111: Using codec to unpack Any type messages to the underlying structure, then we can get proposal message by proposal.Messages without error.

modules/gov/handle_msg.go
```diff
@@ -32,37 +30,35 @@ func (m *Module) HandleMsg(index int, msg sdk.Msg, tx *juno.Tx) error {
...
+		return m.handleVoteEvent(tx, cosmosMsg.Voter, tx.Logs[index].Events)
 	}
 
 	return nil
 }
 
-// handleMsgSubmitProposal allows to properly handle a MsgSubmitProposal
-func (m *Module) handleMsgSubmitProposal(tx *juno.Tx, index int, msg *govtypesv1.MsgSubmitProposal) error {
+// handleSubmitProposalEvent allows to properly handle a handleSubmitProposalEvent
+func (m *Module) handleSubmitProposalEvent(tx *juno.Tx, proposer string, events sdk.StringEvents) error {
```
dadamu, line 57: V1beta1 and V1 messages are sharing the same event, so we can parse event rather than push message directly so that we don't support many version messages later.

https://docs.cosmos.network/main/build/modules/gov#events

modules/gov/utils_events.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+	}
+
+	return 0, fmt.Errorf("no vote option found")
+}
+
+// parseVoteOption returns the vote option from the given string
+// option value in string could be 2 cases, for example:
+// 1. "{\"option\":1,\"weight\":\"1.000000000000000000\"}"
+// 2. "option:VOTE_OPTION_NO weight:\"1.000000000000000000\""
```
dadamu, line 42: VoteOption inside event is represented in these two type format, v1 and v1beta1 currently.

modules/gov/handle_msg.go
```diff
@@ -32,37 +30,35 @@ func (m *Module) HandleMsg(index int, msg sdk.Msg, tx *juno.Tx) error {
...
 
 	case *govtypesv1.MsgDeposit:
-		return m.handleMsgDeposit(tx, cosmosMsg)
+		return m.handleDepositEvent(tx, cosmosMsg.Depositor, tx.Logs[index].Events)
+	case *govtypesv1beta1.MsgDeposit:
+		return m.handleDepositEvent(tx, cosmosMsg.Depositor, tx.Logs[index].Events)
 
 	case *govtypesv1.MsgVote:
-		return m.handleMsgVote(tx, cosmosMsg)
+		return m.handleVoteEvent(tx, cosmosMsg.Voter, tx.Logs[index].Events)
```
dadamu, line 43: Voter event is supported by 0.47.5, we currently better to put Voter from message to make it backward compatible.
https://github.com/cosmos/cosmos-sdk/blob/main/CHANGELOG.md#v0475---2023-09-01

modules/gov/handle_msg.go
```diff
@@ -32,37 +30,35 @@ func (m *Module) HandleMsg(index int, msg sdk.Msg, tx *juno.Tx) error {
 
 	switch cosmosMsg := msg.(type) {
 	case *govtypesv1.MsgSubmitProposal:
-		return m.handleMsgSubmitProposal(tx, index, cosmosMsg)
+		return m.handleSubmitProposalEvent(tx, cosmosMsg.Proposer, tx.Logs[index].Events)
```
dadamu, line 33: We still need put proposer from message since Cosmos-SDK does not support `proposer` attribute inside `submit_proposal` event, it maybe better we post a request to Cosmos-SDK team or just remove this field in the feature. From my side, I would remove it as I would see depositors more than proposer, in addition, proposer must be one of depositors.

modules/gov/handle_msg.go
```diff
@@ -169,37 +170,18 @@ func (m *Module) handleVoteEvent(tx *juno.Tx, voter string, events sdk.StringEve
...
 	if err != nil {
 		return fmt.Errorf("error while saving vote: %s", err)
 	}
 
 	// update tally result for given proposal
-	return m.UpdateProposalTallyResult(msg.ProposalId, tx.Height)
-}
-
-// handleMsgVoteWeighted allows to properly handle a MsgVoteWeighted
-func (m *Module) handleMsgVoteWeighted(tx *juno.Tx, msg *govtypesv1.MsgVoteWeighted) error {
```
dadamu, line 176: It shares the same vote event as MsgVote, so it can be removed safely.

cmd/parse/gov/proposal.go
```diff
@@ -124,7 +125,11 @@ func refreshProposalDetails(parseCtx *parser.Context, proposalID uint64, govModu
...
 	// Handle the MsgSubmitProposal messages
 	for index, msg := range tx.GetMsgs() {
-		if _, ok := msg.(*govtypesv1.MsgSubmitProposal); !ok {
+		_, isMsgSubmitProposalV1 := msg.(*govtypesv1.MsgSubmitProposal)
+		_, isMsgSubmitProposalV1Beta1 := msg.(*govtypesv1beta1.MsgSubmitProposal)
+
+		// Skip if the message is not a submit proposal message
+		if !isMsgSubmitProposalV1 && !isMsgSubmitProposalV1Beta1 {
 			continue
 		}
```
ghost: What about

```go
		switch msg.(type) {
		case *govtypesv1.MsgSubmitProposal, *govtypesv1beta1.MsgSubmitProposal:
			err = govModule.HandleMsg(index, msg, tx)
			if err != nil {
				return fmt.Errorf("error while handling MsgSubmitProposal: %s", err)
			}

		}
```

modules/gov/utils_events.go
ghost: I am still trying to understand cosmos SDK.

We are not able to use provided methods like:

```go
func (m *Vote) GetOptions() []*WeightedVoteOption {
	if m != nil {
		return m.Options
	}
	return nil
}
```

ghost (approved):
Some questions :)

cmd/parse/gov/proposal.go
dadamu: Great, it is cleaner, thanks! https://github.com/forbole/callisto/pull/725/commits/3936cc6693dc527c9ba1ddc0317794902ccbdcab

dadamu:
@0x7u 
>  We are not able to use provided methods like:

Unluckily, the thing I tried is to parse the string from event, and it shows two cases in Cosmos-SDK now.
The option value string is from:
https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/x/gov/types/v1beta1/vote.go#L66
https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/x/gov/types/v1/gov.pb.go#L133

There is no function supporting to parse string into option and also [WeightOptionsFromString](https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/x/gov/types/v1/vote.go#L117) couldn't parse these option string properly since it is used for [cmd string](https://github.com/cosmos/cosmos-sdk/blob/a82615bdf8057bb491539bc4fdd0143174842282/x/gov/client/cli/tx.go#L259).

Therefore we can only parse them by ourselves. If we handle message instance directly, then we need to have both functions, v1 and v1beta1 for each gov messages.

ghost:
> @0x7u
> 
> > We are not able to use provided methods like:
> 
> Unluckily, the thing I tried is to parse the string from event, and it shows two cases in Cosmos-SDK now. The option value string is from: https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/x/gov/types/v1beta1/vote.go#L66 https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/x/gov/types/v1/gov.pb.go#L133
> 
> There is no function supporting to parse string into option and also [WeightOptionsFromString](https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/x/gov/types/v1/vote.go#L117) couldn't parse these option string properly since it is used for [cmd string](https://github.com/cosmos/cosmos-sdk/blob/a82615bdf8057bb491539bc4fdd0143174842282/x/gov/client/cli/tx.go#L259).
> 
> Therefore we can only parse them by ourselves. If we handle message instance directly, then we need to have both functions, v1 and v1beta1 for each gov messages.

Very interesting.

